PR #4, opened 2015-11-25
Conflict with other pods that contain a zip.h file

hkellaway:
I'm experiencing an issue where I have the mParticle pod installed alongside another pod that is using a file named `Zip.h`. My project cannot compile with the two at odds.

Specifically, I have a pod with a dependency on SSZipArchive (https://github.com/ZipArchive/ZipArchive), which is a popular library for zipping. This library has a file named `zip.h`.

Simply changing mParticle's `Zip.h` file name from `MPZip.h` resolved conflicts in my own project - however, I was unable to verify via testing as `pod install` did not work


dcirne:
Thanks for the pull request @hkellaway. We will take a look at it.

dcirne:
@hkellaway in its current format I can't merge the pull request because of other implications in the SDK. I will, however, add a prefix to the class name for the next release, thus avoiding the conflict.

hkellaway:
@dcirne awesome, thanks!

PR #17, opened 2016-03-16
Version 5.4.2 release candidate

dcirne:


Pod/Classes/Network/MPNetworkCommunication.mm
```diff
@@ -39,6 +39,7 @@
 #import "MPDataModelAbstract.h"
 #import "NSUserDefaults+mParticle.h"
 #import "MPSessionHistory.h"
+#import "MPDateFormatter.h"
```
samdozor: can this import and the additional implementations in `MPDateFormatter` be removed/undone, now that 5.4.2 doesn't have the retry-after logic, and we're also not going to be using a date for that header?

PR #23, opened 2016-05-23
use a private suite to prevent the consuming app blowing away mPartic…

ekimia:
…les NSUserDefault keys.


ekimia:
Note this probably is best suited as a reference since I'm sure this will break current consumers of your SDK.

dcirne:
Thank you for the PR @ekimia. We will keep in mind alternative storage for `NSUserDefaults` for a future release of the SDK.

PR #32, opened 2016-10-12
Address sainitizer fix

BrianDoig:
This fixes a simultaneous read and write of the _mpId variable by different threads detected by Xcode 8's AddressSanitizer.  It protects the data by using a serial queue.

### DIFF
--- a/mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
+++ b/mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
@@ -196,11 +196,8 @@ NSString *const kMPCKExpiration = @"e";
 }
 
 - (id)init {
-    // Serial since read/write lock not appropriate given how mpId is created.
-    dispatch_queue_t newQueue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
-    
     self = [super init];
-    if (self && newQueue) {
+    if (self) {
         _consumerInfoId = 0;
     } else {
         self = nil;
@@ -227,8 +224,6 @@ NSString *const kMPCKExpiration = @"e";
 }
 
 - (id)initWithCoder:(NSCoder *)coder {
-    // Serial since read/write lock not appropriate given how mpId is created.
-    dispatch_queue_t newQueue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
     self = [super init];
     if (self) {
         _cookies = [coder decodeObjectForKey:@"cookies"];
--- a/mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
+++ b/mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
@@ -176,6 +176,10 @@ NSString *const kMPCKExpiration = @"e";
 @end
 
 #pragma mark - MPConsumerInfo
+@interface MPConsumerInfo ()
+@property (nonatomic, strong, nonnull) dispatch_queue_t queue;
+@end
+
 @implementation MPConsumerInfo
 
 @synthesize cookies = _cookies;
@@ -183,9 +187,15 @@ NSString *const kMPCKExpiration = @"e";
 @synthesize uniqueIdentifier = _uniqueIdentifier;
 
 - (id)init {
+    // Serial since read/write lock not appropriate given how mpId is created.
+    dispatch_queue_t newQueue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
+    
     self = [super init];
-    if (self) {
+    if (self && newQueue) {
+        _queue = newQueue;
         _consumerInfoId = 0;
+    } else {
+        self = nil;
     }
     
     return self;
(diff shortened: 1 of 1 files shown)


mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
```diff
@@ -176,39 +176,54 @@ - (NSDictionary *)dictionaryRepresentation {
...
 @synthesize mpId = _mpId;
 @synthesize uniqueIdentifier = _uniqueIdentifier;
 
 - (id)init {
+    // Serial since read/write lock not appropriate given how mpId is created.
+    dispatch_queue_t newQueue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
+    
     self = [super init];
-    if (self) {
+    if (self && newQueue) {
```
dcirne: Why does the initialization of the instance depends on creating a queue? Can't the instance be created and the serial queue be lazily created in the getter of the `queue` property?

```
- (dispatch_queue_t)queue {
    if (!_queue) {
        _queue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
    }
    return _queue;
}
```

Can we also rename the property to a more descriptive name such as `cookieQueue` or `accessQueue`?

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
```diff
@@ -176,39 +176,54 @@ - (NSDictionary *)dictionaryRepresentation {
...
+        }
+    });
 }
 
 - (id)initWithCoder:(NSCoder *)coder {
+    // Serial since read/write lock not appropriate given how mpId is created.
+    dispatch_queue_t newQueue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
     self = [super init];
     if (self) {
+        self.queue = newQueue;
```
dcirne: If the implementation creates a queue lazily, we probably wouldn't need to create and assign one over here.

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
```diff
@@ -307,46 +322,52 @@ - (NSDictionary *)localCookiesDictionary {
...
+    dispatch_sync(self.queue, ^{
+        // If we don't have the id, create it.
+        if (!_mpId) {
+            [self willChangeValueForKey:@"mpId"];
+            
+            NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+            NSString *mpIdString = userDefaults[kMPRemoteConfigMPIDKey];
+            
+            if (mpIdString) {
+                [userDefaults removeMPObjectForKey:kMPRemoteConfigMPIDKey];
```
dcirne: Can we reduce the scope of this `dispatch_sync`? The access to `userDefaults` here may need to happen on the main thread.

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
BrianDoig: That's not thread safe.  There is a race condition between checking if the queue is created and creation of the queue.  We could isolate it with a dispatch once however like so.
```
- (dispatch_queue_t)queue {
    static dispatch_queue_t theQueue = nil;
    static dispatch_once_t onceToken;
    dispatch_once(&onceToken, ^{
        theQueue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
    });
    return theQueue;
}
```
This is the method recommended by apple to perform thread safe lazy initialization.  The queue's underlying value can never be accessed from outside of the function, and dispatch once ensures that it's always initialized only once before returning the value.

dcirne: Thank you for the suggestion and explanation @BrianDoig, `dispatch_once` is indeed appropriate for this case. We do utilize it in other parts of the SDK and also in the implementation and initialization of kits.

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
BrianDoig: Unfortunately, I don't think we can.  The test for nil and the creation of the value all need to be inside of one sync block.  

We could remove the access to NSUserDefaults out of the thread, but that means it would always access NSUserDefaults, even if the  value had already been computed.  That would probably be even less efficient.  

On the other hand, this will usually only happen one time and then it will set the value into _mpId.  After that, it will always have a value and the slow path will be skipped.  

Assume that a background thread is currently inside the block, and the main thread is blocked on the generation of this value.  The main thread will then run once the value has been generated, read the value directly and skip over the slow generation code.  There is the slight overhead of the queue emptying and triggering the next one, but that's pretty small and will only happen once.  Once the value has been generated, everything will just become a test that skips over creating the value which is fast.

Even if there 10 more reads in front of the main queue (unlikely), the reads once the value has been created are going to finish extremely fast since they become a single test instruction that jumps to the return statement.

So yes, there is a overhead but it will only paid once, and no matter which thread it took place on, it still had to be paid first before any of the threads could use the value.  The worst case is when multiple threads create a value and then write to the same address at once since that can result in many weird bugs.  Then you have each thread using it's own UUID for the id, then which one you finally settle on is random, and the pointer could even be corrupted since the code for the assignment statement is not atomic.

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
BrianDoig: Yeah, I'm updating it to use the lazy created queue.

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
```diff
@@ -176,34 +176,51 @@ - (NSDictionary *)dictionaryRepresentation {
...
+        theQueue = dispatch_queue_create("com.mParticle.MPCookie.queue", DISPATCH_QUEUE_SERIAL);
+    });
+    return theQueue;
+}
+
 - (id)init {
     self = [super init];
     if (self) {
         _consumerInfoId = 0;
+    } else {
```
dcirne: if `!self` isn't `self` already `nil`?

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
```diff
@@ -307,46 +324,52 @@ - (NSDictionary *)localCookiesDictionary {
...
-    if (MPIsNull(mpId)) {
-        return;
-    }
-    
-    if ([mpId isEqualToNumber:@0]) {
-        _mpId = [self generateMpId];
-    } else {
-        _mpId = mpId;
-    }
+    dispatch_async(self.queue, ^{
```
dcirne: What do you think about using `dispatch_sync` over here?

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
```diff
@@ -307,46 +324,52 @@ - (NSDictionary *)localCookiesDictionary {
...
+    dispatch_sync(self.queue, ^{
+        // If we don't have the id, create it.
+        if (!_mpId) {
+            [self willChangeValueForKey:@"mpId"];
+            
+            NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+            NSString *mpIdString = userDefaults[kMPRemoteConfigMPIDKey];
+            
+            if (mpIdString) {
+                [userDefaults removeMPObjectForKey:kMPRemoteConfigMPIDKey];
```
dcirne: The most critical part as I understand is to execute write NSUserDefaults operations on the main thread (for synchronization issues). In this case, perhaps we can just wrap `removeMPObjectForKey:` on a `dispatch_async` on the main thread:
```
dispatch_async(dispatch_get_main_queue(), ^{
    [userDefaults removeMPObjectForKey:kMPRemoteConfigMPIDKey];
}];
````

mParticle-Apple-SDK/Data Model/MPConsumerInfo.mm
BrianDoig: Change made and checked in for review.

dcirne:
The PR has been accepted and merged to master. The commits have been squashed, thus I'm manually closing this PR.

PR #33, opened 2016-10-25
Remove NSDictionary generic specifications for checkForDeferredDeepLinkWithCompletionHandler callback

suchtomwow:
(See http://useyourloaf.com/blog/using-objective-c-lightweight-generics/ on Objective-C generics)

Because the callback for `checkForDeferredDeepLinkWithCompletionHandler` traces its NSDictionary all the way back to `[Branch getLatestReferringParams]`, which doesn’t specify generics for NSDictionary, there is no guarantee that the keys and values will be of type `<NSString *, NSString *>`. This proves to be an issue when bridging to Swift, as there are times when `[Branch getLatestreferringParams]` _can_ return a dictionary of type `[NSObject : AnyObject]`, which causes a crash at runtime because Swift is expecting `[String : String]`, but is getting `[String : NSNumber]`.

Therefore, the callback can only be as specific as its upstream dependencies, which is to say, not at all. Thus, removing the <NSString *, NSString *> generic specification on the callback.

Removing the specification results in bridging to `[NSObject : AnyObject]` when consumed in Swift, which requires the consumer to cast from `AnyObject` to the expected type (`String` in this case) when pulling values out of the dictionary.

**This PR has implications for the mparticle-integrations/mparticle-apple-integration-branchmetrics repo implementation, for which there is a PR already created [here](https://github.com/mparticle-integrations/mparticle-apple-integration-branchmetrics/pull/4).**

Here is how I’ve implemented the contract change in our code:
```
MParticle.sharedInstance().checkForDeferredDeepLinkWithCompletionHandler { linkInfo, error in
    if let canonicalURL = linkInfo?["$canonical_url"] as? String, let url = NSURL(string: canonicalURL) {
        navigateToLink(url)
    }
}
```

### DIFF
--- a/mParticle-Apple-SDK/Kits/MPKitProtocol.h
+++ b/mParticle-Apple-SDK/Kits/MPKitProtocol.h
@@ -60,7 +60,7 @@
 - (void)deinit;
 
 #pragma mark Application
-- (nonnull MPKitExecStatus *)checkForDeferredDeepLinkWithCompletionHandler:(void(^ _Nonnull)(NSDictionary<NSString *, NSString *> * _Nullable linkInfo, NSError * _Nullable error))completionHandler;
+- (nonnull MPKitExecStatus *)checkForDeferredDeepLinkWithCompletionHandler:(void(^ _Nonnull)(NSDictionary * _Nullable linkInfo, NSError * _Nullable error))completionHandler;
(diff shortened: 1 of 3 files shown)


dcirne:
Thanks for the PR @suchtomwow. We will take a look at it.

PR #37, opened 2016-12-21
stop using subscript operatator on NSUserDefaults.  

navisingh:


dcirne:
@navisingh, Thank you for the PR!

We are looking into this for the future. However, since this update touches core functionality of important storage of data point, will require a new unit test for every value it touches, and significant review, we can not accept the PR as is at this time.

PR #52, opened 2017-09-28
Set `CLANG_ENABLE_CODE_COVERAGE` to `NO`

mattbanach:
Per Apple: “Do not submit apps with LLVM profiling instrumentation or
coverage collection enabled. Turn off LLVM profiling or code coverage,
rebuild your app and resubmit the app.”


samdozor:
@mattbanach thanks for the PR. To confirm - are you using Carthage or including our project as a subproject? AFAIK this change has no affect on Cocoapods as our project file isn't involved there.

samdozor:
Ah - looks like a rampant issue with Xcode 9 and Carthage: https://github.com/Carthage/Carthage/issues/2056

We'll get this merged and assess affected kits ASAP.

mattbanach:
Yep, correct! Using Xcode + Carthage surfaces the issue.

Issue #56, opened 2017-11-20
AppsFlyer is not getting currency for commerce events


alexandra-mikh:
Hello,
Original issue we are facing is that currency property which is passed to mParticle is not sent to AppsFlyer for commerce events.
Checking out the mparticle-apple-sdk it turns out that at some point the property is lost while copying MPCommerceEvent inside mparticle sdk. 
`__block MPCommerceEvent *forwardCommerceEvent = [commerceEvent copy];`
Checking out the MPCommerceEvent NSCopying protocol implementation, you can see that currency property is not being copied, which later is affecting AppsFlyer event setup
`if (commerceEvent.currency) {
            values[AFEventParamCurrency] = commerceEvent.currency;
  }`
I'm not sure if it was done for a reason though. 
![screen shot 2017-11-20 at 15 01 55](https://user-images.githubusercontent.com/5527326/33027960-70a7e6ce-ce0c-11e7-84a3-0100ade11a85.png)

samdozor:
Hi @alexandra-mikh this fix has been released in SDK version 6.15.18. Thanks for your patience!

alexandra-mikh:
Hi Sam, 

Thanks for cooperation and letting us know.

Kind regards, 
Alexandra.

PR #57, opened 2017-11-20
copy currency property in MPCommerceEvent

alexandra-mikh:
Please, find more details in the issue.
https://github.com/mParticle/mparticle-apple-sdk/issues/56

MPCommerce event losing a currency property after being copied.


samdozor:
Hi @alexandra-mikh thanks for reporting this issue and creating this PR! We'll review and if approved include it in our next release.

neharikasingh13:
Hi @samdozor thanks! Can we get an idea of when we can expect this to be released?

Thanks,
Neharika

Issue #59, opened 2018-01-15
AppsFlyer commerce event wrong quantity parameter


alexandra-mikh:
Hello,
I was working with marketing team on the commerce events and we've found out a weird issue. AppsFlyer commerce events' quantity is set to random value when 1 is expected, e.g. 15, 14, 2, 5, etc. 
![image](https://user-images.githubusercontent.com/5527326/34947431-237fed7a-fa1b-11e7-8009-e0b67ac5585c.png)

Checking out actual implementation I've discovered the following: whenever appsflyer commerce event is being composed, quantity param is taken from the MPCommerceEvent object (MPKitAppsFlyer.m line 273). 
![screen shot 2018-01-15 at 17 01 19](https://user-images.githubusercontent.com/5527326/34945838-c0afed44-fa15-11e7-9cb1-4c7965948c4f.png)
However, in the MPCommerceEvent implementation you can see that count is taken from userDefinedAttributes, not productList as I would expect. 
![screen shot 2018-01-15 at 17 05 16](https://user-images.githubusercontent.com/5527326/34946038-6087e1f0-fa16-11e7-93df-fab7d9b8caa6.png)
Can you please let me know what's the expected behaviour in this case? Currently we assume that quantity in this context is a products count. 
Thanks.

samdozor:
hey @alexandra-mikh agreed this is definitely a bug. We'll have a look at your PR and get this addressed for our next release.

alexandra-mikh:
Hi @samdozor, thanks for the quick reply. Can you please us know when do you have the next release planned?

samdozor:
This has been resolved in mParticle-AppsFlyer 6.16.1 - thanks for the report!

PR #60, opened 2018-01-15
MPCommerceEvent count depends on productsList

alexandra-mikh:
This is PR following the [issue](https://github.com/mParticle/mparticle-apple-sdk/issues/59).
In brief, whenever commerce event for AppsFlyer is composed, af_quantity value is set to MPCommerceEvent count, which is based on userDefinedAttributes. 
I'm not sure it's right, but we expect there a product count, so in this PR I'm setting count based on productsList count. 
Please, let me know if my understanding is wrong and af_quantity stands for something else. 
Thanks.


samdozor:
@alexandra-mikh thanks for this PR. Your PR will only take the length of the array into account, rather than the total quantity of each product in the array. We've implemented a slightly different approach directly in the AppsFlyer kit, which you can see below. We've just released mParticle-AppsFlyer 6.16.1 to Cocoapods with that change:

https://github.com/mparticle-integrations/mparticle-apple-integration-appsflyer/commit/11b1f9b4ad8723a336cb481cd2f7a1dfc8889ba1

PR #64, opened 2018-04-06
Forward userAttribute registration instead of removing it.

jervine10:
## Summary
When registering user attributes without a value, the attribute is not forwarded to any kits. Instead `removeUserAttribute` is called. I've updated to call `setUserAttribute:value:`

### DIFF
--- a/mParticle-Apple-SDK/Identity/MParticleUser.m
+++ b/mParticle-Apple-SDK/Identity/MParticleUser.m
@@ -169,11 +169,11 @@
                                                                                     [kit setUserAttribute:key value:value];
                                                                                 }];
                                        } else {
-                                           [[MPKitContainer sharedInstance] forwardSDKCall:@selector(removeUserAttribute:)
+                                           [[MPKitContainer sharedInstance] forwardSDKCall:@selector(setUserAttribute:value:)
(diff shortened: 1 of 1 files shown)


jervine10:
Thanks for getting back to me about this. I did try using setUserTag but the Urban Airship kit didn’t seem to respond to that. That’s why I looked into this method. As soon as I modified it I was able to receive push notifications with the provided attribute.

jervine10:
@peterjenkins 
Seems like this is more of a problem with `MPKitUrbanAirship` than with the mParticle SDK. It does **not** implement the optional method `setUserTag`. After looking through it some more I don't see a way to remove any of the tags that get registered either. In our use case, users will be adding and removing tags quite frequently. Without the ability to remove they'll receive push notifications that aren't relevant anymore. 

Clearly this pull request is not relevant, but there does seem to be an issue somewhere. Who is responsible for the Urban Airship Kit? Is it mParticle or Urban Airship?

jervine10:
@peterjenkins Thanks for the quick work on this. I'll pull in the update and let you know how it goes.

jervine10:
@peterjenkins 
I got this working with your new update, thank you.

PR #81, opened 2018-11-05
Add imports for category definitions

shepting:
## Summary
There are two files that are missing the direct import for the category definitions that they create. This allows those files to be built if the top-level header is not directly added.

## Testing Plan
Built into the Airbnb Alpha app.

## Master Issue
No formal issue created. Discussion on mParticle Slack channel only.


shepting:
@peterjenkins Thanks! I'd still build locally to make sure these were right. I typed them manually into the web UI based on our local changes.

PR #116, opened 2020-10-09
Change imports to modular

KieranLafferty:
## Summary
Using the the quotations for imports conflicts with other sqlite pods that may be included from other dependencies causing a 

sqlite.h file not found

## Testing Plan
Build and run and observe a successful build result


BrandonStalnaker (approved):
Sorry the delay @KieranLafferty . This looks great, I just wanted to explore if we could transfer more than sqlite to modular.

PR #117, opened 2020-10-29
Use Default Endpoint for Upload Header

BrandonStalnaker:


BrandonStalnaker:
https://git.corp.mparticle.com/mParticle/mparticle-apple-sdk/pull/649

PR #127, opened 2021-05-08
Annotate `MParticleOptions` and `MPIdentityApiRequest` factory initializers with `instancetype`

djs-code:
## Summary
When mparticle-apple-sdk is pulled in via Cocoapods as a static library w/ `use_modular_headers!` set, the factory initializers for `MParticleOptions` and `MPIdentityApiRequest` (and likely others, though these are the two that affect us) don't get bridged as expected, resulting in the following error:

Screen Shot 2021-05-07 at 9 34 14 PM

This change replaces the return types of these two initializers with `instancetype`, which corrects the bridging behavior.

## Testing Plan
Tested via compiling with our app.  This is generally a compile-time enhancement, and shouldn't have any run-time differences.

### DIFF
--- a/mParticle-Apple-SDK/Include/MPIdentityApiRequest.h
+++ b/mParticle-Apple-SDK/Include/MPIdentityApiRequest.h
@@ -14,8 +14,8 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface MPIdentityApiRequest : NSObject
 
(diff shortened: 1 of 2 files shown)


djs-code:
Thanks @peterjenkins!  I have another very similar PR up on **mparticle-apple-integration-button** if you're up for it!

https://github.com/mparticle-integrations/mparticle-apple-integration-button/pull/13

PR #129, opened 2021-06-01
Update setATTState to setATTStatus

suemparticle:
## Summary
Our public docsite correctly uses setATTStatus. However, this migration guide is not up to date.


BrandonStalnaker (approved):
LGTM I'll ask the others how they'd like to handle merging this in

PR #134, opened 2021-09-30
fix: Expose Session Start Time

alla-uday:
On mParticle's Web and Android SDK's it's possible to retrieve the current sessions start time:

MParticle.getInstance()?.currentSession?.sessionStartTime
Web - mParticle.Store.sessionStartDate.getTime()

Currently on the iOS SDK MPSession is internal only while MParticleSession is what is provided to users (which doesn't expose the session start time property).


BrandonStalnaker (approved):
LGTM

PR #140, opened 2022-01-06
Fix: Correct IDFA Enrichment when Changed during Session

BrandonStalnaker:
## Summary
Basically the device Info that's used for the event batch comes from the MPSession saved to the local database. So that saved version of the deviceInfo wasn't getting updated until a new session is created (restart). ATT and IDFA were implemented before we began [caching app and device info](https://github.com/mParticle/mparticle-apple-sdk-internal/pull/17/files#diff-4166039127eb7c02c467286fe092e331049fc718b8770abc7ffcf78fe061b8d5R159) so it been broken in this way since that change.

## Testing Plan
Tested using sample App

## Master Issue
Closes https://go.mparticle.com/work/76333

### DIFF
--- a/UnitTests/MPUploadBuilderTests.m
+++ b/UnitTests/MPUploadBuilderTests.m
@@ -9,6 +9,7 @@
 #import "MPIntegrationAttributes.h"
 #import "MPPersistenceController.h"
 #import "MPBaseTestCase.h"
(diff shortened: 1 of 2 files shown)


peterjenkins (approved):
LGTM

Issue #143, opened 2022-03-25
MPAppDelegateProxy does not forward protocol conformance


cbaltzer:
We have a React Native app that uses mParticle in combination with [react-native-app-auth](https://github.com/FormidableLabs/react-native-app-auth). Implementing this involves adding a property and a protocol to our AppDelegate. 

`MPAppDelegateProxy` is intercepting calls to the app delegate, which is normally no problem because it forwards all selectors. However, it appears that RN checks `conformsToProtocol` before calling the selector, which `MPAppDelegateProxy` does not implement. The flow works if we add the necessary protocol to `MPAppDelegateProxy` directly, but that's not a workable solution as it gets overwritten any time we do a pod install (in CI/CD builds for example).  

PR: #144

einsteinx2:
This fix has just gone out with release v8.8.0: https://github.com/mParticle/mparticle-apple-sdk/releases/tag/v8.8.0

PR #144, opened 2022-03-25
Adds conformsToProtocol forwarding to MPAppDelegateProxy

cbaltzer:
## Summary
Adds an implementation of `conformsToProtocol:` to `MPAppDelegateProxy` that bubbles up protocol implementations from the base AppDelegate. Issue is described in #143.

Had to add an extra singleton in order to get the original app delegate types. If anyone can think of a better way, input is extremely welcome. 

## Testing Plan
Manually smoke tested in our setup, and unit tested. Should be low-to-no impact for other integrations, but unknown.

## Master Issue
Implementation as described in [#143](https://github.com/mParticle/mparticle-apple-sdk/issues/143)

### DIFF
--- a/mParticle-Apple-SDK/AppNotifications/MPAppDelegateProxy.m
+++ b/mParticle-Apple-SDK/AppNotifications/MPAppDelegateProxy.m
@@ -13,9 +13,18 @@
 
 @implementation MPAppDelegateProxy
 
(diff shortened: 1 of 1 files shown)


BrandonStalnaker (approved):
LGTM

PR #147, opened 2022-04-26
fix: update active kits header

willpassidomo:
## Summary
The "x-mp-kits" header field was previously sending a list of kits that were actively receiving events. Now it will send a list of kits that have been configured, even if they are not actively receiving events due to the SDK state. For Example, a kit that was shut down due to the current user's logged-in status not matching its configurations directive previous would not have been included in this list but now will be included

## Testing Plan
Modified a few existing tests to include checks for the new `[kjitContainer configuredKits]` method where the results would differ from `[kitContainer activeKits]`. Also included a new unit test with a more focused scenario. Updated a networking tests to make sure the `configuredKits` method was properly being applied to `x-mp-kits`

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-3738

### Related (Android) PR
https://github.com/mParticle/mparticle-android-sdk/pull/126

### DIFF
--- a/mParticle-Apple-SDK/Kits/MPKitContainer.mm
+++ b/mParticle-Apple-SDK/Kits/MPKitContainer.mm
@@ -1893,7 +1893,7 @@ static NSMutableSet <id<MPExtensionKitProtocol>> *kitsRegistry;
 }
 
 - (nullable NSArray<NSNumber *> *)configuredKitsRegistry {
-    BOOL anyKitsIncluded = self.supportedKits != nil && self.supportedKits.count > 0;
+    BOOL anyKitsIncluded = self.supportedKits.count > 0;
     BOOL anyKitsConfigured = self.kitConfigurations != nil && self.kitConfigurations.count > 0;
     if (!anyKitsIncluded || !anyKitsConfigured) {
         return nil;
--- a/mParticle-Apple-SDK/Kits/MPKitContainer.mm
+++ b/mParticle-Apple-SDK/Kits/MPKitContainer.mm
@@ -1894,7 +1894,7 @@ static NSMutableSet <id<MPExtensionKitProtocol>> *kitsRegistry;
 
 - (nullable NSArray<NSNumber *> *)configuredKitsRegistry {
     BOOL anyKitsIncluded = self.supportedKits.count > 0;
-    BOOL anyKitsConfigured = self.kitConfigurations != nil && self.kitConfigurations.count > 0;
+    BOOL anyKitsConfigured = self.kitConfigurations.count > 0;
     if (!anyKitsIncluded || !anyKitsConfigured) {
         return nil;
     }
--- a/mParticle-Apple-SDK/Kits/MPKitContainer.mm
+++ b/mParticle-Apple-SDK/Kits/MPKitContainer.mm
@@ -1892,6 +1892,16 @@ static NSMutableSet <id<MPExtensionKitProtocol>> *kitsRegistry;
     completionHandler(projectedEvents, appliedProjections);
 }
 
+- (nonnull NSArray<NSNumber *> *)configuredKitsRegistry {
+    NSMutableArray<NSNumber *> *configuredKits = [[NSMutableArray alloc] initWithCapacity:self.kitConfigurations.count];
+    for (NSNumber *kitId in self.kitConfigurations.allKeys) {
+        if ([self.supportedKits containsObject:kitId]) {
+            [configuredKits addObject:kitId];
+        }
+    }
+    return configuredKits;
+}
+
 #pragma mark Public methods
 - (nullable NSArray<id<MPExtensionKitProtocol>> *)activeKitsRegistry {
     if (kitsRegistry.count == 0) {
--- a/mParticle-Apple-SDK/Kits/MPKitContainer.mm
+++ b/mParticle-Apple-SDK/Kits/MPKitContainer.mm
@@ -1892,7 +1892,12 @@ static NSMutableSet <id<MPExtensionKitProtocol>> *kitsRegistry;
     completionHandler(projectedEvents, appliedProjections);
 }
 
-- (nonnull NSArray<NSNumber *> *)configuredKitsRegistry {
+- (nullable NSArray<NSNumber *> *)configuredKitsRegistry {
+    BOOL anyKitsIncluded = self.supportedKits != nil && self.supportedKits.count > 0;
+    BOOL anyKitsConfigured = self.kitConfigurations != nil && self.kitConfigurations.count > 0;
+    if (!anyKitsIncluded || !anyKitsConfigured) {
+        return nil;
+    }
     NSMutableArray<NSNumber *> *configuredKits = [[NSMutableArray alloc] initWithCapacity:self.kitConfigurations.count];
     for (NSNumber *kitId in self.kitConfigurations.allKeys) {
         if ([self.supportedKits containsObject:kitId]) {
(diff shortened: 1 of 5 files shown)


mParticle-Apple-SDK/Kits/MPKitContainer.mm
```diff
@@ -1892,6 +1892,16 @@ - (void)project:(id<MPExtensionKitProtocol>)kitRegister event:(MPEvent *const)ev
     completionHandler(projectedEvents, appliedProjections);
 }
 
+- (nonnull NSArray<NSNumber *> *)configuredKitsRegistry {
```
einsteinx2, line 1895: What's the reason for having this nonnull and returning an empty array vs what `activeKitsRegistry` does which is return nil if there's no kits? I think we should stay consistent unless there's a good reason, so in this case that would mean standardizing on what `activeKitsRegistry` is doing.

mParticle-Apple-SDK/Network/MPURLRequestBuilder.m
```diff
@@ -170,16 +170,9 @@ - (NSMutableURLRequest *)build {
...
-            if (activeKitsRegistry.count > 0) {
-                NSMutableArray<NSNumber *> *activeKitIds = [[NSMutableArray alloc] initWithCapacity:activeKitsRegistry.count];
-                
-                for (id<MPExtensionKitProtocol> kitRegister in activeKitsRegistry) {
-                    [activeKitIds addObject:kitRegister.code];
-                }
-                
-                kits = [activeKitIds componentsJoinedByString:@","];
-            }
+            NSMutableArray<NSNumber *> *configuredKitIds = [[NSMutableArray alloc] initWithArray:[[MParticle sharedInstance].kitContainer configuredKitsRegistry]];
```
einsteinx2, line 173: Nitpick, but for consistency with the rest of this file and general readability, I'd prefer to see dot syntax here. 

So `[MParticle sharedInstance].kitContainer` would be `MParticle.sharedInstance.kitContainer`

einsteinx2: Also this code can be slightly simplified to use `mutableCopy` instead of the whole alloc/init.

So like this: `[MParticle.sharedInstance.kitContainer mutableCopy];`

einsteinx2: Though be aware that would mean if `kitContainer` or `configuredKitsRegistry` is nil, `configuredKitIds` would be nil, and so `kits` would be nil, whereas this way, `configuredKitIds` will always be an object, so you may not want to use `mutableCopy` if that's required.

einsteinx2 (changes requested):

mParticle-Apple-SDK/Kits/MPKitContainer.mm
willpassidomo: So, I did notice this too, and went a bit back and forth on how to balance consistency vs eveything else.. Maybe this is more philosophical, but I tend to think it's almost always better to remove nullability/nillibility from the equation if you reasonably can?

I'll update it though because consistency is very important and I'm not trying to make style changes to the repo in this little PR :)

mParticle-Apple-SDK/Network/MPURLRequestBuilder.m
willpassidomo: 🙌 will do

willpassidomo: for my own knowledge, when can you use dot syntax instead of `[]`? I feel like sometimes I try to use the dot syntax and it's fine, but sometimes the XCode linter throws a fit..

mParticle-Apple-SDK/Network/MPURLRequestBuilder.m
```diff
@@ -170,7 +170,7 @@ - (NSMutableURLRequest *)build {
                 kits = nil;
             }
             
-            NSMutableArray<NSNumber *> *configuredKitIds = [[NSMutableArray alloc] initWithArray:[[MParticle sharedInstance].kitContainer configuredKitsRegistry]];
+            NSArray<NSNumber *> *configuredKitIds = MParticle.sharedInstance.kitContainer.configuredKitsRegistry.copy;
```
einsteinx2, line 173: Ahh I didn't realize you didn't actually need a mutable array, and all you're doing is converting it to a string anyway. In that case, there's no need for the copy. This line should just be `NSArray<NSNumber *> *configuredKitIds = MParticle.sharedInstance.kitContainer.configuredKitsRegistry;`

Alternatively, you could remove this line altogether and replace line 175 with `kits = [MParticle.sharedInstance.kitContainer.configuredKitsRegistry componentsJoinedByString:@","];`

einsteinx2 (changes requested):

mParticle-Apple-SDK/Network/MPURLRequestBuilder.m
willpassidomo: sweet, gotcha...I didn't realize either until I went back to refactor :)

mParticle-Apple-SDK/Network/MPURLRequestBuilder.m
einsteinx2: In general, it should also be possible to use dot syntax anywhere you use bracket syntax, except for when the method takes parameters. I'd have to see a concrete example of where the linter is giving problems.

mParticle-Apple-SDK/Kits/MPKitContainer.mm
einsteinx2: Yeah agreed on the general concept of removing nullability, but also agree that consistency is important, so for not lets stay consistent, and later we can refactor the SDK in general to remove nullability.

mParticle-Apple-SDK/Kits/MPKitContainer.mm
```diff
@@ -1892,6 +1892,21 @@ - (void)project:(id<MPExtensionKitProtocol>)kitRegister event:(MPEvent *const)ev
     completionHandler(projectedEvents, appliedProjections);
 }
 
+- (nullable NSArray<NSNumber *> *)configuredKitsRegistry {
+    BOOL anyKitsIncluded = self.supportedKits != nil && self.supportedKits.count > 0;
```
einsteinx2, line 1896: Sorry for the nonstop nitpicks haha, but the nil checks are unnecessary here. In Objective-C, it's completely valid to send messages (call methods) on nil objects. If they have an object return type, they will return nil, and if they have an integer return type will return 0, etc. 

So in this case, `self.supportedKits.count` will evaluate to 0 if `supportedKits` is nil, so the single check of `BOOL anyKitsIncluded = self.supportedKits.count > 0;` is valid in all cases, even if `supportedKits` is nil.

Same goes for the next line as well, it's cleaner and totally valid to just write: `BOOL self.kitConfigurations.count > 0;`

einsteinx2 (changes requested):

mParticle-Apple-SDK/Kits/MPKitContainer.mm
willpassidomo: dude, no problem at all. I'd rather learn it hear than go through life writing sub-par obj-c :)

willpassidomo:
@einsteinx2 bump :)

einsteinx2 (approved):
Fixed the other line I mentioned in my last comment rather than ping-ponging back to you again for something so trivial. Looks good to me now, I have no more nitpicks left haha.

PR #148, opened 2022-05-03
fix: Fix Dereferenced Garbage Pointer

BrandonStalnaker:
## Summary
A pattern was  being used where a static NSString is converted to a std::string via the UTF8String property of NSString, stored in a std::vector, and then retrieved via operator[]. This result is then copied into a new string at the call site, which is where the app crashes.

## Master Issue
Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-3658?atlOrigin=eyJpIjoiY2Q0NmJiNTY1NzAyNGE0MzgxM2RmZGE0MjI4MzViMTciLCJwIjoiaiJ9


mParticle-Apple-SDK/Ecommerce/MPCommerceEvent.mm
```diff
@@ -263,8 +261,8 @@ - (NSMutableArray *)productsList {
 
 #pragma mark Private methods
 - (void)setEventType {
-    static const vector<EventType> productActionEventType {AddToCart, RemoveFromCart, AddToWishlist, RemoveFromWishlist, Checkout, CheckoutOption, Click, ViewDetail, Purchase, Refund};
-    static const vector<EventType> promotionActionEventType {PromotionClick, PromotionView};
+    static const vector<MPEventType> productActionEventType {MPEventTypeAddToCart, MPEventTypeRemoveFromCart, MPEventTypeAddToWishlist, MPEventTypeRemoveFromWishlist, MPEventTypeCheckout, MPEventTypeCheckoutOption, MPEventTypeClick, MPEventTypeViewDetail, MPEventTypePurchase, MPEventTypeRefund};
```
einsteinx2, line 264: We should be able to get rid of this use of std::vector here as well, looks like it's pointless as we just immediately cast back to Obj-C enum type.

einsteinx2:
I'm going to give this a deeper look later, am slammed rest of today.

mParticle-Apple-SDK/Ecommerce/MPCommerceEvent.mm
BrandonStalnaker: It still requires a fair bit of casting but at least its all Objective C now.

einsteinx2 (approved):
Changes look great, so nice to see how much cleaner the code is without all that unnecessary C++. Since we're not longer doing the thing that was causing the crash, hopefully this takes care of it.

I was going to spend some time and do some more C++ refactoring, but honestly I have no spare time right now, so I'll go ahead and approve this since that's really unrelated anyway.

PR #152, opened 2022-05-18
Remove Undefined Hasher Behavior

BrandonStalnaker:
## Summary
Replaced signed that have undefined behavior with more appropriate unsigned types

## Testing Plan
Our HasherTest Unit tests continued to succeed with this change and no unexpected behavior was experienced while testing with the sample app

## Master Issue
Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-3659


mParticle-Apple-SDK/Custom Modules/MPCustomModulePreference.m
```diff
@@ -135,10 +135,10 @@ - (NSString *)defaultValueForMacroPlaceholder:(NSString *)macroPlaceholder __att
         CFUUIDRef uuidRef = CFUUIDCreate(kCFAllocatorDefault);
         CFUUIDBytes uuidBytes = CFUUIDGetUUIDBytes(uuidRef);
         
-        SInt64 lsbBytes[8] = {uuidBytes.byte8, uuidBytes.byte9, uuidBytes.byte10, uuidBytes.byte11,
+        uint64_t lsbBytes[8] = {uuidBytes.byte8, uuidBytes.byte9, uuidBytes.byte10, uuidBytes.byte11,
                               uuidBytes.byte12, uuidBytes.byte13, uuidBytes.byte14, uuidBytes.byte15};
         
-        SInt64 value = 0;
+        uint64_t value = 0;
```
peterjenkins, line 141: Correct me if I'm wrong, but I think this going to change behavior if you don't cast back into a signed integer afterwards.

mParticle-Apple-SDK/Utils/MPHasher.cpp
```diff
@@ -11,7 +10,7 @@ namespace mParticle {
         string lowerCaseStringToHash = stringToHash;
         transform(lowerCaseStringToHash.begin(), lowerCaseStringToHash.end(), lowerCaseStringToHash.begin(), ::tolower);
         
-        int hash = 0;
+        unsigned int hash = 0;
```
einsteinx2, line 13: I know this was original just `int`, but since we're changing the type anyway, I'd like to see a `uintXX_t` type just so we're explicit about the size. Easy way to find would be to add a log statement printing the sizeof int to confirm it then use that bit size.

Also, if it doesn't hurt anything, maybe we can specify the return type as well, but if that's too invasive (also modifying various other places, then we can skip that and clean this up more later).

einsteinx2: Actually let's look into this kind of thing later when we do a bigger refactor.

PR #155, opened 2022-07-15
fix: Add filtering check to legacy code path

peterjenkins:
## Summary
Add filtering check to legacy code path

## Testing Plan
Manual testing

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4319


einsteinx2 (approved):
This is good to merge, we tested together and it works as expected

PR #157, opened 2022-08-01
feat: match android type support for custom attributes

BrandonStalnaker:
## Summary
We discussed this in stand up and determined it makes sense to have iOS match Android behavior in terms of serializing dictionaries when sending to the server but leave the objects as-is when passing to kits. 

## Testing Plan
Updated unit test to include cases with dictionaries

## Master Issue
Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-4087

### DIFF
--- a/UnitTests/MPEventTests.mm
+++ b/UnitTests/MPEventTests.mm
@@ -47,7 +47,8 @@
     }
     
     NSDictionary *eventInfo = @{@"speed":@25,
-                                @"modality":@"sprinting"};
+                                @"modality":@"sprinting",
+                                @"stats":@{}};
     
     event.customAttributes = eventInfo;
     event.category = @"Olympic Games";
@@ -86,8 +87,8 @@
     XCTAssertNotNil(event.customAttributes, @"Should not have been nil.");
     XCTAssertNotNil(event.info, @"Should not have been nil.");
 
-    XCTAssertEqual(event.customAttributes.count, 2, @"Should have been two values in the customAttributes dictionary.");
-    XCTAssertEqual(event.info.count, 2, @"Should have been two values in the info dictionary.");
+    XCTAssertEqual(event.customAttributes.count, 3, @"Should have been three values in the customAttributes dictionary.");
+    XCTAssertEqual(event.info.count, 3, @"Should have been three values in the info dictionary.");
 
     NSDictionary *copyEventInfo = [eventInfo copy];
     event.customAttributes = copyEventInfo;
@@ -163,7 +164,8 @@
     MPEvent *event = [[MPEvent alloc] initWithName:@"Dinosaur Run" type:MPEventTypeOther];
     event.duration = eventDuration;
     event.customAttributes = @{@"speed":@25,
-                   @"modality":@"sprinting"};
+                               @"modality":@"sprinting",
+                               @"stats":@{}};
     event.category = @"Olympic Games";
     
     [session incrementCounter];
@@ -180,6 +182,7 @@
     
     NSDictionary *attributes = @{@"speed":@25,
                                  @"modality":@"sprinting",
+                                 @"stats":@{},
                                  @"$Category":@"Olympic Games",
                                  @"EventLength":eventDuration};
     XCTAssertEqualObjects(dictionaryRepresentation[kMPAttributesKey], attributes, @"Attributes are not being set correctly.");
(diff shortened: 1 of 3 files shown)


mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.m
```diff
@@ -72,8 +72,7 @@ - (id)valueForCaseInsensitiveKey:(NSString *)key {
             transformedDictionary[key] = [[NSString alloc] initWithData:obj encoding:NSUTF8StringEncoding];
         } else {
             MPILogError(@"Data type is not supported as an attribute value: %@ - %@", obj, [[obj class] description]);
```
BrandonStalnaker, line 74: Question for reviewers: Should I keep this warning in?

UnitTests/MPEventTests.mm
```diff
@@ -163,7 +164,8 @@ - (void)testDictionaryRepresentation {
     MPEvent *event = [[MPEvent alloc] initWithName:@"Dinosaur Run" type:MPEventTypeOther];
     event.duration = eventDuration;
     event.customAttributes = @{@"speed":@25,
-                   @"modality":@"sprinting"};
+                               @"modality":@"sprinting",
+                               @"stats":@{}};
```
einsteinx2, line 168: Same comment re: other dictionary test cases

UnitTests/MPEventTests.mm
```diff
@@ -180,6 +182,7 @@ - (void)testDictionaryRepresentation {
     
     NSDictionary *attributes = @{@"speed":@25,
                                  @"modality":@"sprinting",
+                                 @"stats":@{},
```
einsteinx2, line 185: Same comment re: other dictionary test cases

mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.m
einsteinx2: Seems like we should add another if else case for dictionaries?

mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.m
```diff
@@ -72,8 +72,7 @@ - (id)valueForCaseInsensitiveKey:(NSString *)key {
...
-            NSAssert([obj isKindOfClass:[NSString class]], @"Data type is not supported as an attribute value");
-            return;
+            transformedDictionary[key] = [obj description];
```
einsteinx2, line 75: I remember we decided on this (using `description`) since that's what Android is doing, though this will result in a different string than Android. I think it would be ideal to use a JSON string on both platforms, but I think we decided that was out of scope for this work and could be something we do in the future...

einsteinx2 (dismissed):

mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.m
BrandonStalnaker: I actually did this initially but was checking count (example below) which I didn't realize was what was making me not hit that statement. Thanks for making me take another look.

```
 else if ([obj isKindOfClass:[NSDictionary class]] && [(NSDictionary *)obj count] > 0) {
            transformedDictionary[key] = [obj description];
        } else if ([obj isKindOfClass:[NSMutableDictionary class]] && [(NSDictionary *)obj count] > 0) {
            transformedDictionary[key] = [obj description];
        } else {
            MPILogError(@"Data type is not supported as an attribute value: %@ - %@", obj, [[obj class] description]);
            NSAssert([obj isKindOfClass:[NSString class]], @"Data type is not supported as an attribute value");
            return;
        }
        ```

File: UnitTests/MPEventTests.mm [hunk 5]
```diff
@@ -47,7 +47,8 @@ - (void)testInstance {
     }
     
     NSDictionary *eventInfo = @{@"speed":@25,
-                                @"modality":@"sprinting"};
+                                @"modality":@"sprinting",
+                                @"stats":@{}};
```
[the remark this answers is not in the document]
BrandonStalnaker: Added a test case for dictionary with values and one with dictionary values in the dictionary

peterjenkins (approved):

einsteinx2 (approved):
LGTM

PR #163, opened 2022-10-11
fix: Allow null event attributes

BrandonStalnaker:
## Summary
This is an inconsistency between data planning and iOS. Android support this without issue. Based on those two inconsistencies, I’m rationalizing this as a bug.

Context: In JSON schema, there is a difference between null and missing, and therefore in our data planning feature, we allow you to specific that event attributes can be null. Steps to repro:

Create a data plan that with an event containing an event attribute value that is both both required and to allowed to be null (note, this is different than optional/not-required, you need to literally specify that the value is allowed to be nullable)

With the Apple SDK, create an event with nsnull as an event attribute value

when the SDK uploads, in live stream and in the data planning violation report, it’ll say that a required attribute is missing

## Testing Plan
Tested with live stream and added unit tests

## Master Issue
Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-4273

### DIFF
--- a/.github/workflows/cross-platform-tests.yml
+++ b/.github/workflows/cross-platform-tests.yml
@@ -11,14 +11,8 @@ on:
 jobs:
   run-ios-tests:
     timeout-minutes: 30
-    runs-on: macos-latest
+    runs-on: macos-12
     steps:
-      - name: "Update XCode tools"
-        run: |
-          sudo xcode-select -v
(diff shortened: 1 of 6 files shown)


peterjenkins (approved):

einsteinx2 (approved):
LGTM

PR #168, opened 2022-11-17
fix: Remove unnecessary calls to remove observer

peterjenkins:
## Summary
Remove calls to removeObserver no longer necessary now that we no longer support iOS 8

## Testing Plan
Validated per Apple [docs](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1413994-removeobserver) that this is no longer necessary since observer will be removed automatically.

## Master Issue
Closes https://go.mparticle.com/work/SQDSDKS-4830


einsteinx2 (approved):
LGTM

PR #171, opened 2022-11-29
feat: Remove MPSegment code

BrandonStalnaker:
## Summary
While I've not been able to document an example of it leaking, it has been reported to have the potential to leak. So since the code is also deprecated and unused it made more sense to just go ahead and remove it.

## Testing Plan
Ran full unit test and spent the day smoke testing in a sample app
 
## Master Issue
Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-4835


einsteinx2 (approved):
LGTM

PR #180, opened 2023-01-30
fix: Correct location flag issue when publishing

BrandonStalnaker:
## Summary
  - These commits give two options for the issue where clients attempts at publishing are triggering the NSLocationAlwaysAndWhenInUsageDescription flag from Apple. The first commit expands the pre processor flag to a CLLocation call that was not included before. The second commit removes all uses of CLLocation. When reviewing please give your opinion on which is the preferred solution.

 ## Testing Plan
 - Both commits tested on simulator to confirm no crash or issue in that respect

 - Closes https://go.mparticle.com/work/SQDSDKS-4997

### DIFF
--- a/mParticle-Apple-SDK/Include/mParticle.h
+++ b/mParticle-Apple-SDK/Include/mParticle.h
@@ -26,7 +26,9 @@
 #import <UIKit/UIKit.h>
 
 #if TARGET_OS_IOS == 1
-    #import <CoreLocation/CoreLocation.h>
+    #ifndef MPARTICLE_LOCATION_DISABLE
+        #import <CoreLocation/CoreLocation.h>
+    #endif
     #import <WebKit/WebKit.h>
 #endif
 
@@ -980,6 +982,7 @@ Defaults to false. Prevents the eventsHost above from overwriting the alias endp
  */
 @property (nonatomic) BOOL backgroundLocationTracking;
 
+#ifndef MPARTICLE_LOCATION_DISABLE
 /**
  Gets/Sets the current location of the active session.
  @see beginLocationTracking:minDistance:
@@ -1015,6 +1018,7 @@ Defaults to false. Prevents the eventsHost above from overwriting the alias endp
  Ends geographic location tracking.
  */
 - (void)endLocationTracking;
+#endif
 #endif
 
 #pragma mark - Network Performance
--- a/mParticle-Apple-SDK/Utils/MPCurrentState.m
+++ b/mParticle-Apple-SDK/Utils/MPCurrentState.m
@@ -205,7 +205,7 @@ NSString *const kMPStateFreeDiskSpaceKey = @"fds";
     }
     return _statusBarOrientation;
 }
-#endif
+#endif // TARGET_OS_IOS
 
 #pragma mark Public instance methods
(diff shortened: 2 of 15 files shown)


einsteinx2:
So looking over this, I think the solution should be somewhere in the middle. Basically if you take the second commit, but instead of deleting all the lines, take each deleted portion (including the entire `MPLocationManager.h/m` files) and wrap each in `#ifndef MPARTICLE_LOCATION_DISABLE`. So basically the same approach we're currently taking, but more aggressive to completely remove the location tracking code when `MPARTICLE_LOCATION_DISABLE` is specified.

einsteinx2:
I think the problem the customer is having is related to Apple being more aggressive in it's auto-detection and probably tripping up on some other bit of code we're not removing. So if we just use the existing flag to really remove everything location related, we should be good.

mParticle-Apple-SDK/Include/mParticle.h
```diff
@@ -980,6 +982,7 @@ Defaults to false. Prevents the eventsHost above from overwriting the alias endp
  */
```
einsteinx2, line 982: Should we move the ifndef to above this comment (right under the `#if TARGET_OS_IOS = 1` line) to exclude this option entirely? Or would that cause other problems? I assume right now it just won't do anything if location is disabled?

mParticle-Apple-SDK/Utils/MPCurrentState.m
```diff
@@ -188,7 +190,10 @@ - (NSNumber *)deviceOrientation {
 }
 
 - (NSNumber *)gpsState {
-    BOOL gpsState = [CLLocationManager authorizationStatus] && [CLLocationManager locationServicesEnabled];
+    BOOL gpsState = false;
```
einsteinx2, line 193:
```suggestion
    BOOL gpsState = NO;
```
Just for consistency in Obj-C this should probably be `NO` instead of `false`.

mParticle-Apple-SDK/mParticle.m
```diff
@@ -1344,15 +1348,24 @@ - (void)kitInstance:(NSNumber *)kitCode completionHandler:(void (^)(id _Nullable
...
+#endif
 }
 
 - (void)setBackgroundLocationTracking:(BOOL)backgroundLocationTracking {
     [MPListenerController.sharedInstance onAPICalled:_cmd parameter1:@(backgroundLocationTracking)];
     
+#ifndef MPARTICLE_LOCATION_DISABLE
     [MParticle sharedInstance].stateMachine.locationManager.backgroundLocationTracking = backgroundLocationTracking;
+#else
+    MPILogDebug(@"Automatic background tracking has been disabled to support users excluding location services from their applications.");
```
einsteinx2, line 1364: Ahh ok I see what you did. We can probably just leave this option in then.

einsteinx2 (changes requested):

einsteinx2 (approved):
Tested locally and confirmed complete removal of all CoreLocation related code in a test app using `strings` tool. This should be good to go. Not sure what else we could possibly remove at this point to make Apple any happier...

eseay:
@einsteinx2 I noticed in this change that you updated the README to include instructions about how to set this flag using CocoaPods. Is there a way to make this work if we integration mParticle via SPM?

PR #186, opened 2023-03-09
feat: Port first Obj-C file to Swift

einsteinx2:
## Summary
To begin using Swift in the SDK, we need to first port one file to Swift, then update our release action to support binary releases for SPM (since SPM doesn't support Obj-C and Swift in the same framework unless it's distributed as a binary). This is the first step toward that goal. This is meant to be merged into the dedicated feature branch for this, then once we update the release action and do regression testing, it will be merged into development/main.

 ## Testing Plan
- Added unit tests before porting the file to Swift and confirmed they passed before and after the porting.

 - Closes https://go.mparticle.com/work/SQDSDKS-5139


einsteinx2:
Note this only has a ❌ because we’re merging into a feature branch instead of development, but all tests passed.

BrandonStalnaker (approved):
LGTM

PR #187, opened 2023-03-24
feat: Use Apple Search Ads for Attribution

BrandonStalnaker:
## Summary
 - Implement Apple Search Ads for Attribution in way similar to our existing attribution support for ADClient

 ## Testing Plan
 - Tested locally on simulator

 - Closes https://go.mparticle.com/work/SQDSDKS-5122

### DIFF
--- a/UnitTests/MParticleOptionsTests.m
+++ b/UnitTests/MParticleOptionsTests.m
@@ -66,14 +66,14 @@
 - (void)testSetSearchAdsAttributionDefault {
     _options = [MParticleOptions optionsWithKey:@"unit_test_app_key" secret:@"unit_test_secret"];
     
-    XCTAssertTrue(_options.collectSearchAdsAttribution, @"Search ads attribution should be collected by default");
+    XCTAssertFalse(_options.collectSearchAdsAttribution, @"Search ads attribution shouldn't be collected by default");
 }
 
 - (void)testSetSearchAdsAttributionSet {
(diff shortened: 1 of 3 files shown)


samdozor:
haven't done a full review yet but some early thoughts:

- I think this should be an opt-in thing
- make sure you're not making http requests on the main thread :-)
- potentially we shouldn't be doing this at all, and actually we should be grabbing just the token, adding it to the batch, and then sending it to partners

for that last item - what we need to do is search the integrations codebase to see which use the current object, and then look at the partner's docs to see what they are expecting these days - maybe they just want the token.

BrandonStalnaker:
@samdozor 
- For opt in, would you want the default to be active with the option to disable? We currently have a field in mParticleOptions where it defaults to true `collectSearchAdsAttribution`
- Good on the async
- Looking at the server code it looks like its expected to be in the format Dictionary<string, Dictionary<string, string>> where I assume the outer key to be the version as we currently send. https://git.corp.mparticle.com/mParticle/mPServer/blob/661f1dc91fd844df858a3465f0396cd28ac2bebe/src/Sdk/MP.Sdk.ClientDto/SdkMessages/AppInfo.cs#L193

BrandonStalnaker:
Talked to @samdozor and finalized a few decisions. 1. We changed add attribution functionality to opt in. 2. When collecting the attribution with the token we convert the format of the data to the prior format so that no backend changes are necessary. So this is ready to review @einsteinx2 and/or Sam.

einsteinx2 (approved):
Review the solution live over Zoom

PR #197, opened 2023-06-06
feat: New architecture standalone items

einsteinx2:
## Summary
 - Starting to add the new interfaces
 - Some from the ticket were not included as they already exist on iOS (unlike Web). 
 - We already have MPMessage, but I added a new class called `BaseMPMessage` for now to match Android which we can discuss later.
 - Functions that were using async/await in Kotlin were rewritten to use callbacks. We can add typealiases later to give the callback types names.

 ## Testing Plan
 - No testing, these are just interfaces and empty classes

 - Closes https://go.mparticle.com/work/SQDSDKS-5379

### DIFF
--- a/mParticle-Apple-SDK/Next Gen SDK/MParticleDataRepository.swift
+++ b/mParticle-Apple-SDK/Next Gen SDK/MParticleDataRepository.swift
@@ -1,0 +1,16 @@
+//
+//  MParticleDataRepository.swift
+//  mParticle-Apple-SDK
+//
+//  Created by Ben Baron on 6/5/23.
+//
+
+import Foundation
+
+internal class MParticleDataRepository {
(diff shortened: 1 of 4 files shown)


mParticle-Apple-SDK/Next Gen SDK/MParticleDataRepository.swift
```diff
@@ -0,0 +1,16 @@
...
+//  MParticleDataRepository.swift
+//  mParticle-Apple-SDK
+//
+//  Created by Ben Baron on 6/5/23.
+//
+
+import Foundation
+
+internal class MParticleDataRepository {
+    func updateSession(data: BaseMPMessage, completionHandler: (() -> Void)? = nil) -> Void {
```
alexs-mparticle, line 11: Is `completionHandler` something that needs to be in every function param?

markvdouw (approved):

Issue #191, opened 2023-05-03
GitHub Actions Hanging


eseay:
After updating our mParticle version to v8.13.0, we noticed that GitHub Actions CI jobs are hanging indefinitely - in particular, our `xcodebuild test` invocations. They get to the end of package resolution and just sit there.

I noticed that in [this commit](https://github.com/mParticle/mparticle-apple-sdk/commit/e482df0c9496d47e73276655afd72a4f483b2d66), you updated mParticle to leverage binary targets instead of source targets, and my hypothesis is that this is related to the issue.

By pinning explicitly to v8.12.0 (pre-binary targets), the issue went away.

Last week, when we integrated Lottie into our app, we noticed the same exact behavior. This is discussed in [this issue](https://github.com/airbnb/lottie-ios/issues/2051) on the Lottie repository.

There is one striking similarity. Lottie suggests depending on [lottie-spm](https://github.com/airbnb/lottie-spm/blob/main/Package.swift) to get the benefits of binary targets, so we did that. After integrating it, we saw our actions hanging. Per the suggestion of their engineers, we tried to build from source instead via [lottie-ios](https://github.com/airbnb/lottie-ios/blob/master/Package.swift), and we noticed that the problem went away.

So the request isn't necessarily to solve this problem at its root for my team, but perhaps would it be possible to offer the option of building from source in addition to binary targets until we can determine the root cause? We're pretty aggressive abound updating our mParticle version to always be on the latest, so I'd really prefer not to pin to an outdated version for a long time.

einsteinx2:
Thank you for the feedback, this is very valuable to know. Unfortunately, SPM currently does not support source targets for mixed Objective-C and Swift targets. That's why we moved to binary targets. Cocoapods does support building from source though, so that's an option, though that's obviously a bigger change since it forces you into their xcworkspace format...

Could you provide a minimum test project that reproduces the CI issue? I'd like to look into it on our end. Maybe a stripped down version of your project and workflow that you can share?

einsteinx2:
Also fwiw, some of our popular partner SDKs used with our kits, such as Braze, also use binary targets, though our Braze kit itself is currently all Objective-C and so is a source target, but does have Braze's binary target as a dependency in SPM. We hadn't heard of any issues from customers regarding CI issues using our Braze kit, so I wonder what's different about this case that's causing the issue.

Also does `xcodebuild test` work on your local machine? That is to say, is this a CI only issue?

[...]

PR #199, opened 2023-06-13
feat: Upload build artifacts to S3 for SPM

einsteinx2:
## Summary
 - Upload build artifacts used by SPM to an S3 bucket to work around an Xcode bug that unnecessarily requires credentials to download artifacts from the Github release which breaks CI for some users.

 ## Testing Plan
- Tested each piece independently (tested uploading to this S3 bucket using these credentials, tested running the release script but uploading to a test bucket on a fork to show that the env variables would work inside the release script)
- Can't do a full end to end test until our next release of the SDK, but everything should work, and if we run into issues, changes should be minimal as all pieces have been validated. 

 - Closes https://go.mparticle.com/work/SQDSDKS-5416

### DIFF
--- a/.github/workflows/sdk-release.yml
+++ b/.github/workflows/sdk-release.yml
@@ -43,9 +43,16 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Validate environment
+        env:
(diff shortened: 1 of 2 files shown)


einsteinx2:
close https://github.com/mParticle/mparticle-apple-sdk/issues/191

BrandonStalnaker (approved):
LGTM

Issue #204, opened 2023-07-05
Latest v8.14.0  release crash on startWithOptions


abdela7ad:
We run into this crash and I think this happens because of this [RR](https://github.com/mParticle/mparticle-apple-sdk/commit/f870b2075ba587827a3fc0e4b0e0602f869ea83d)

it happens immediately after we call `startWithOptions`

`*** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM removeObjectsInRange:]: range {1, 1} extends beyond bounds [0 .. 0]'
*** First throw call stack:
(0x18cdd4cb4 0x185e743d0 0x18cf7af84 0x18cde318c 0x10b4fe78c 0x10b502e30 0x18cdde69c 0x18ce670f0 0x10b502bc8 0x18cdde69c 0x18ce670f0 0x10b502a6c 0x18cdde69c 0x18ce670f0 0x10b50292c 0x18cdde69c 0x18ce670f0 0x10b502500 0x10b50bdc8 0x10b4ec450 0x10b5017d0 0x10b50bc2c 0x10b50ba78 0x10b50ac38 0x10c02c520 0x10c02e038 0x10c0360b0 0x10c036df4 0x10c043c74 0x1ecb54ddc 0x1ecb54b7c)
libc++abi: terminating due to uncaught exception of type NSException`

<img width="554" alt="Screenshot 2023-07-05 at 8 42 28 PM" src="https://github.com/mParticle/mparticle-apple-sdk/assets/32923534/76ba20ef-5359-4622-9ddb-77318e2de0de">

einsteinx2:
@abdela7ad Thank you for the report! I'm having trouble reproducing this on my end using that same SDK version and our CI isn't hitting it either. Can you please provide any additional details on reproduction steps? Thanks!

In the meantime I'll open an internal bug ticket to track this.

einsteinx2:
Tracking internally as https://go.mparticle.com/work/SQDSDKS-5537

einsteinx2:
@abdela7ad The fix has been released as version v8.14.1. Please update and it should resolve your problem. If it persists, please reopen this issue. Thanks again for the report!

PR #205, opened 2023-07-06
fix: crash in userIdentitiesForUserId with multiple invalid identity types

einsteinx2:
## Summary
The SDK could crash on start if multiple invalid user identity types are saved (e.g. `MPIdentityIOSAdvertiserId` and `MPIdentityPushToken`).

Issue reported here: https://github.com/mParticle/mparticle-apple-sdk/issues/204

 ## Testing Plan
New unit tests were written to confirm the crash, then the logic was fixed and tests confirmed no crash and correct logic.

 - Closes https://go.mparticle.com/work/SQDSDKS-5537


BrandonStalnaker (approved):
LGTM

PR #210, opened 2023-07-20
feat: Begin implementing “MP as a kit” and next gen architecture

einsteinx2:
## Summary
 - Begin implementing “MP as a kit” and first classes from the next gen architecture

 ## Testing Plan
 - Unit tests were updated to the new architecture and all tests pass
 - End to end testing using a test app

 - Closes https://go.mparticle.com/work/SQDSDKS-5509


mParticle-Apple-SDK/Identity/MPIdentityApi.m
```diff
@@ -469,8 +470,7 @@ - (BOOL)aliasUsers:(MPAliasRequest *)aliasRequest {
     }
     
     dispatch_async([MParticle messageQueue], ^{
-        
-        [MParticle.sharedInstance.backendController skipNextUpload];
+        [(MParticleKit *)[MParticle.sharedInstance kitInstance:[MParticleKit kitCode]] skipNextUpload];
```
alexs-mparticle, line 473: What is the purpose of `skipNextUpload`? We don't have a similar concept in Web at the moment, and if this is new, what is the intention here?

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
```diff
@@ -0,0 +1,50 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MPEventLogging : NSObject <MPComponent>
+
+@property (nonatomic, strong, nullable) NSMutableSet<MPEvent *> *eventSet;
+
+- (void)leaveBreadcrumb:(nonnull MPEvent *)event completionHandler:(void (^ _Nonnull)(MPEvent * _Nonnull event, MPExecStatus execStatus))completionHandler;
+- (void)logError:(nullable NSString *)message exception:(nullable NSException *)exception topmostContext:(nullable id)topmostContext eventInfo:(nullable NSDictionary *)eventInfo completionHandler:(void (^ _Nonnull)(NSString * _Nullable message, MPExecStatus execStatus))completionHandler;
+- (void)logCrash:(nullable NSString *)message stackTrace:(nullable NSString *)stackTrace plCrashReport:(nonnull NSString *)plCrashReport completionHandler:(void (^ _Nonnull)(NSString * _Nullable message, MPExecStatus execStatus)) completionHandler;
+- (void)logBaseEvent:(nonnull MPBaseEvent *)event completionHandler:(void (^ _Nonnull)(MPBaseEvent * _Nonnull event, MPExecStatus execStatus))completionHandler;
```
alexs-mparticle, line 41: Your `logEvent` and `logBaseEvent` signatures look identical. Is there a major difference in the iOS SDK between `MPEvent` and `MPBaseEvent`?

mParticle-Apple-SDK/MPBackendController.h
```diff
@@ -79,18 +63,9 @@ extern const NSInteger kInvalidKey;
 - (nonnull MParticleSession *)tempSession;
 - (void)beginTimedEvent:(nonnull MPEvent *)event completionHandler:(void (^ _Nonnull)(MPEvent * _Nonnull event, MPExecStatus execStatus))completionHandler;
 + (BOOL)checkAttribute:(nonnull NSDictionary *)attributesDictionary key:(nonnull NSString *)key value:(nonnull id)value error:(out NSError *__autoreleasing _Nullable * _Nullable)error;
-- (nullable MPEvent *)eventWithName:(nonnull NSString *)eventName;
 - (nullable NSString *)execStatusDescription:(MPExecStatus)execStatus;
 - (nullable NSNumber *)incrementSessionAttribute:(nonnull MPSession *)session key:(nonnull NSString *)key byValue:(nonnull NSNumber *)value;
 - (nullable NSNumber *)incrementUserAttribute:(nonnull NSString *)key byValue:(nonnull NSNumber *)value;
-- (void)leaveBreadcrumb:(nonnull MPEvent *)event completionHandler:(void (^ _Nonnull)(MPEvent * _Nonnull event, MPExecStatus execStatus))completionHandler;
```
markvdouw, line 86: Great I would love to se highly cohesive implementation with SRP

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
```diff
@@ -0,0 +1,50 @@
...
+};
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MPEventLogging : NSObject <MPComponent>
+
+@property (nonatomic, strong, nullable) NSMutableSet<MPEvent *> *eventSet;
+
+- (void)leaveBreadcrumb:(nonnull MPEvent *)event completionHandler:(void (^ _Nonnull)(MPEvent * _Nonnull event, MPExecStatus execStatus))completionHandler;
+- (void)logError:(nullable NSString *)message exception:(nullable NSException *)exception topmostContext:(nullable id)topmostContext eventInfo:(nullable NSDictionary *)eventInfo completionHandler:(void (^ _Nonnull)(NSString * _Nullable message, MPExecStatus execStatus))completionHandler;
```
markvdouw, line 39: Haven't we assumed that we were going to collapse error and exception (in your case "crash") on a single function?

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
markvdouw: I think we should go for logEvent. If by any means let say we need to handle differently the event types (and eventually cast it or something), that shouldn't be exposed in implementation or naming, and should be transparent for the user. If you ask me, Im not a fan of MPBaseEvent either (which we also have on Android), our hierarchy and architecture shouldn't be exponsed.

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
```diff
@@ -0,0 +1,50 @@
...
+- (void)logCrash:(nullable NSString *)message stackTrace:(nullable NSString *)stackTrace plCrashReport:(nonnull NSString *)plCrashReport completionHandler:(void (^ _Nonnull)(NSString * _Nullable message, MPExecStatus execStatus)) completionHandler;
+- (void)logBaseEvent:(nonnull MPBaseEvent *)event completionHandler:(void (^ _Nonnull)(MPBaseEvent * _Nonnull event, MPExecStatus execStatus))completionHandler;
+- (void)logEvent:(nonnull MPEvent *)event completionHandler:(void (^ _Nonnull)(MPEvent * _Nonnull event, MPExecStatus execStatus))completionHandler;
+- (void)logCommerceEvent:(nonnull MPCommerceEvent *)commerceEvent completionHandler:(void (^ _Nonnull)(MPCommerceEvent * _Nonnull commerceEvent, MPExecStatus execStatus))completionHandler;
+- (void)logNetworkPerformanceMeasurement:(nonnull MPNetworkPerformance *)networkPerformance completionHandler:(void (^ _Nullable)(MPNetworkPerformance * _Nonnull networkPerformance, MPExecStatus execStatus))completionHandler;
```
markvdouw, line 44: Please @einsteinx2 correct me if Im wrong, but I see that we are passing callbacks. Are we planning to leave the callback implementation for backwards compatibility and go to an async implementation (like we are on other platforms with suspend functions, promises, futures or not?) - just a question

mParticle-Apple-SDK/Next Gen/MPEventLogging.m
```diff
@@ -0,0 +1,353 @@
...
+#import "MParticleWebView.h"
+#import "MPDevice.h"
+
+@interface MParticle()
+@property (nonatomic, strong, nonnull) MPBackendController *backendController;
+@property (nonatomic, strong) MPStateMachine *stateMachine;
+@property (nonatomic, strong) MPPersistenceController *persistenceController;
+@end
+
+@implementation MPEventLogging
```
markvdouw, line 50: The implementation itself is kind of a mixture between an adapter (old/new interfaces and architecture) and a delegate right?

If we want to iterate over this in time to create extra layer to achieve high cohesivness and almost single responsibility of the implementation classes we might want to create an adapter and make this call the adapter class which will actually have the business logic.
That way it might be easier to iterate (just an idea)

logEvent() {
     //add params based on new interface implementation and pass it to the adapter function. 
     adapter.logEvent()
 }

On top of it.. the MPEventLogging should be acting almost as a visitor in the registered kits? 
The two prev concepts match if we were implement this as:

logEvent() {
    mediator.getKits.forEach {. kit ->
         kit.logEvent() // by doing this each kit will have a concrete impl and the mPKit have an adapter internally
    }
}

mParticle-Apple-SDK/Next Gen/MPMediator.m
```diff
@@ -0,0 +1,23 @@
...
+//  Created by Ben Baron on 6/29/23.
+//
+
+#import "MPMediator.h"
+
+@interface MPMediator()
+@property (nonatomic, strong) MPEventLogging *eventLogging;
+@end
+
+@implementation MPMediator
```
markvdouw, line 14: Shouldn't kit be registered here and this serve as a glue for the components to interact and visit the kits?

mParticle-Apple-SDK/Next Gen/MParticleKit/MParticleKit.m
```diff
@@ -0,0 +1,192 @@
...
+#import "MParticleWebView.h"
+#import "MPMessage.h"
+#import "MPPersistenceController.h"
+#import "MPStateMachine.h"
+#import "MPUploadBuilder.h"
+#import "MPBackendController.h"
+
+const NSInteger kMParticleKitCode = 999999999;
+
+@interface MParticle()
```
markvdouw, line 23: Im not sure if this is an implementation outline in progress or the finished version?
I think that the MParticle should almost serve as a facade and maybe have a state. I wouldn't include a low level layer as the persistance in the highest external one, or things like dataPlanId/version (maybe we can create a manager for those)

mParticle-Apple-SDK/Kits/MPKitContainer.mm
```diff
@@ -241,6 +242,25 @@ - (void)freeKit:(NSNumber *)integrationId {
     }
 }
 
+- (void)registerMParticleKit {
```
markvdouw, line 245: registerMParticleKit means register any class implementing the MParticleKitProtocol or specifically the mP kit.
I thing that having an mP kit should give us the ability treat them exactly the same as interchangeable, therefore I would use agnostic names (which in this case I dont understand if we are referring to a concrete implementation or the abstraction)

markvdouw (changes requested):

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
einsteinx2: @alexs-mparticle MPEvent is a subclass of MPBaseEvent and internally MPEvents are logged via logBaseEvent. However see below regarding changes to these methods in this work:

@markvdouw For the scope of this work, we're not changing existing method signatures or business logic, only moving them. Later when we do more work toward the new API, we'll condense these into the simpler methods we discussed before (not having so many different methods to log events and adding a shim from the existing methods to the new, cleaner API).

mParticle-Apple-SDK/MPBackendController.h
einsteinx2: This method is not actually removed, only moved.

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
einsteinx2: See my previous comment, that change will come later and is outside of scope for this initial work.

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
einsteinx2: Again, no method signatures are being changed in this initial work, only the object graph (mediator/eventLogging/etc)

To directly answer the question though, on iOS until we can target iOS 13 as a minimum we can't use async/await at all, so we have to use callbacks, not just as a fallback. Later when we can increase the minimum target, we'll be able to switch to async/await and either remove the callbacks entirely or more likely deprecate them and leave them until the next major version or keep them as a fallback if necessary for Obj-C users.

mParticle-Apple-SDK/Next Gen/MPEventLogging.m
einsteinx2: MPEventLogging at this point just groups the existing event logging methods so they can be refactored and shimmed/adapted later. I really wanted to reduce the scope of the initial architecture changes as much as possible.

mParticle-Apple-SDK/Next Gen/MPMediator.m
einsteinx2: I had been thinking of the mediator as just the owner of the various components, but we could certainly move the loading of the mP Kit there if it makes sense. For this initial work, let's leave it as-is, but I'm about to start the spec to continue the work on the mP Kit, so we can include that there if it makes sense.

mParticle-Apple-SDK/Next Gen/MParticleKit/MParticleKit.m
einsteinx2: Agreed in the medium/long term, but for this initial change, I was trying to limit the scope as much as possible so I didn't want to make big changes to the mParticle class yet. I actually had pulled a bunch of logic out of it and into a separate class, but it just made the PR harder to grok as it added more noise, so I pulled it out and left it in a side branch for now.

mParticle-Apple-SDK/Kits/MPKitContainer.mm
einsteinx2: In this case it refers to registering the actual MParticleKit class. My goal with having a separate registration method was to keep it distinct from user provided sideloaded kits. Later we may have other hard coded kits that would also have their own register methods, and this method may be optionally called instead of always called.

mParticle-Apple-SDK/Next Gen/MPEventLogging.h
markvdouw: Great

markvdouw (approved):

BrandonStalnaker (approved):

PR #211, opened 2023-07-28
fix: Update Imports For React Native

BrandonStalnaker:
## Summary
 - Imports were not working with Cocoapods without the 'use_frameworks!` directive which also prevented usage with React Native which doesn't support that directive without a bunch of workarounds.

 ## Testing Plan
 - Testing across CocaoPods, SPM, No-Location/Locations builds, and React Native Projects

 - Closes https://go.mparticle.com/work/SQDSDKS-4646

### DIFF
--- a/mParticle-Apple-SDK/Swift.h
+++ b/mParticle-Apple-SDK/Swift.h
@@ -5,12 +5,10 @@
 //  Created by Ben Baron on 3/24/23.
 //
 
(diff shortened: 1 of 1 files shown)


einsteinx2:
I pushed a new branch called `fix/5564-Update-Imports-For-React-Native` to make CI happy

PR #212, opened 2023-07-28
fix: Update imports for react native compatibility

BrandonStalnaker:
## Summary
 - React Native was having trouble with the imports and in the process of correcting the imports for that we realized 'use_frameworks!' was no longer required for initializing with cocaopods.

 ## Testing Plan
 - Tested for React Native, Cocaopods, and SPM for Location and No-Location builds

 - Closes https://go.mparticle.com/work/SQDSDKS-5564


einsteinx2 (approved):
End to end tested this thoroughly with Brandon to confirm this works with all CoocaPods and SPM configurations and that it works correctly in a React Native test project without the `use_frameworks!` directive.

PR #216, opened 2023-08-29
test: Update string check for Xcode 15 compat

einsteinx2:
## Summary
We were previously directly checking the internal obj-c string class name in 2 unit tests. This has changed to the internal swift class name in iOS 17/Xcode 15. I modified the checks to use the more generic `isKindOfClass:` and now they work on both iOS 17 and earlier versions.

 ## Testing Plan
 - Tests confirmed using both Xcode 14 and 15

 - Closes https://go.mparticle.com/work/SQDSDKS-5711


BrandonStalnaker (approved):
LGTM

PR #218, opened 2023-08-29
test: Handle different URL formatting in iOS 17

einsteinx2:
## Summary
This test just confirms that if a customer sets their custom hosts incorrectly by prepending `https://`, the URL will be incorrect. In iOS 17 it's still incorrect, just in a slightly different way. This won't affect customers with currently correct configurations, so I've just adjusted the test to handle both formats.

 ## Testing Plan
 - Confirmed it works on both iOS 16 and 17

 - Closes https://go.mparticle.com/work/SQDSDKS-5712


BrandonStalnaker (approved):
LGTM

PR #221, opened 2023-08-30
feat: Remove Disk Space API Usage

BrandonStalnaker:
## Summary
 - Remove CurrentState entirely as part of removing all uses of the Disk Space API

 ## Testing Plan
 - Tested locally and end to end with the Livestream. One of the main code paths were this data was being included is when logging identity changes so I confirmed these events along with a broad range of others still ends up on the live stream correctly. 
 
Screenshot 2023-08-30 at 10 14 19 AM
Screenshot 2023-08-30 at 10 14 07 AM

 - Closes https://go.mparticle.com/work/SQDSDKS-5547


einsteinx2 (approved):
lgtm

PR #223, opened 2023-09-26
feat: Direct routing to correct pod

einsteinx2:
## Summary
 - Directly route to the correct pod using the api key prefix when feature flag is enabled

 ## Testing Plan
 - Added unit tests
 - E2E tested against all pods

 - Closes https://go.mparticle.com/work/SQDSDKS-5771

### DIFF
--- a/mParticle-Apple-SDK/Network/MPNetworkCommunication.m
+++ b/mParticle-Apple-SDK/Network/MPNetworkCommunication.m
@@ -49,9 +49,9 @@ NSString *const kMPIdentityURL = @"";
 NSString *const kMPIdentityKey = @"identity";
 
 NSString *const kMPURLScheme = @"https";
-NSString *const kMPURLHost = @"nativesdks.mparticle.com";
 NSString *const kMPURLHostConfig = @"config2.mparticle.com";
-NSString *const kMPURLHostIdentity = @"identity.mparticle.com";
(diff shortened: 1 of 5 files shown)


BrandonStalnaker (approved):
LGTM

mParticle-Apple-SDK/Network/MPNetworkCommunication.m
```diff
@@ -114,6 +114,31 @@ - (instancetype)init {
 }
 
 #pragma mark Private accessors
+
+- (NSString *)defaultHostWithSubdomain:(NSString *)subdomain apiKey:(NSString *)apiKey enableDirectRouting:(BOOL)enableDirectRouting {
+    if (enableDirectRouting) {
+        NSArray *splitKey = [apiKey componentsSeparatedByString:@"-"];
```
markvdouw, line 120: What would happen if the subdomian doesnt have the - for splitting?

markvdouw (approved):

mParticle-Apple-SDK/Network/MPNetworkCommunication.m
einsteinx2: if the API key doesn't have a dash (aka it's an old key without a prefix) this will just return an array with a single value, which I check for on the next line.

PR #228, opened 2023-10-09
feat: 5343 Refactor Kit Filter Hash Functions

BrandonStalnaker:
## Summary
 - Currently the code that hashes each value and compares those hashes to the ones from the kit config is scattered all over the codebase. Since we’ll need to reuse that code in the MPSideloadedKit class to implement the “add filter” methods, we should first refactor it into a static utility class.

 ## Testing Plan
 - Tested locally and with new unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-5343


einsteinx2 (approved):
Looks great! Thanks for adding all the tests!

PR #229, opened 2023-10-10
feat: Use direct url routing config feature flag

einsteinx2:
## Summary
 - Reads the feature flag from the server config to decide whether to enable direct url routing or not (defaults to off if the key isn't included in the config)

 ## Testing Plan
 - Tested using a QA environment with the feature flag present as well as tested against production without the feature flag to confirm it defaults to false
 - Confirmed that new installs will grab the config and have access to the feature flag before making more API calls

 - Closes https://go.mparticle.com/work/SQDSDKS-5777


BrandonStalnaker (approved):
LGTM

PR #231, opened 2023-10-17
feat: Begin Implementation MPSideloadedKit Filtering Methods

BrandonStalnaker:
## Summary
 - The work on the MPSideloadedKit class as specified in the tech spec will be broken into 2 parts. This is the first part which just includes creating the class and the “add filter” methods.

 ## Testing Plan
 - Tested locally

 - Closes https://go.mparticle.com/work/SQDSDKS-5344


einsteinx2 (approved):
lgtm

PR #233, opened 2023-10-24
chore: Remove MPHasher C++ code and reduce number of ObjC++ files

einsteinx2:
## Summary
 - Remove MPHasher C++ code and reduce number of ObjC++ files

 ## Testing Plan
 - All unit tests show the same hashes being generated
 - Smoke tested using local test app

 - Closes https://go.mparticle.com/work/SQDSDKS-5906


einsteinx2:
Note that this seems like a big PR but there are actually relatively few real changes.

PR #240, opened 2023-11-27
fix: CFNetwork upload warning setting body twice

einsteinx2:
## Summary
 - Starting with Xcode 15/iOS 17, there is a new warning in CFNetwork if you call `[NSURLSession uploadTaskWithRequest:fromData:]` if the request passed in already has its body set. In this case, we set the same data as the `HTTPBody` property of the request that we pass in to that method, so in the end the call worked fine, but now warns that we're setting it twice. Since we always set the post body on the request when we call `[MPURLRequestBuilder build]` on line 246, I've switched to using the more correct `[NSURLSession dataTaskWithRequest]` which just uses the HTTPBody from the request instead of the data we pass in. This works exactly the same, but fixes the warning.

 ## Testing Plan
- E2E tested using a test app

 - Closes https://go.mparticle.com/work/SQDSDKS-5921

### DIFF
--- a/.github/workflows/native-tests.yml
+++ b/.github/workflows/native-tests.yml
@@ -16,10 +16,8 @@ jobs:
         platform: [iOS, tvOS]
         scheme: [mParticle-Apple-SDK, mParticle-Apple-SDK-NoLocation]
         include:
-          - xcode: "15.0"
-            os: "17.0"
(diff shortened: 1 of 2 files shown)


BrandonStalnaker (approved):
LGTM

PR #241, opened 2023-11-29
test: Fix unit tests and add isLaunch test

einsteinx2:
## Summary
I found that there was an MPBackgroundController unit test called testPreviousForegroundTime that was failing only on Github due to the way it was implemented. While investigating that, I found that I had the isLaunch boolean backwards, and it should also have its own unit test.

I also noticed that some tests were disabled for tvOS when there's no reason for that as the behavior should be the same, so I enabled them.

I also fixed how some of the other tests were implemented regarding messageQueue threading. 

 ## Testing Plan
- Confirmed via breakpoints when running the previousForgroundTime test that the assertions are hitting correctly now
- Tested isLaunch behavior in a test app locally and new unit test added
- Confirmed all tests run on tvOS as well

 - Closes https://go.mparticle.com/work/SQDSDKS-5981

### DIFF
--- a/UnitTests/MPBackendControllerTests.m
+++ b/UnitTests/MPBackendControllerTests.m
@@ -732,15 +732,10 @@
         }];
         [expectation fulfill];
     });
-    XCTestExpectation *expectation2 = [self expectationWithDescription:@"dispatch_after expectation"];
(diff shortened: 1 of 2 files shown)


samdozor (approved):

BrandonStalnaker (approved):
LGTM

PR #246, opened 2024-01-16
fix: Rare arithmetic overflow crash in MPIHasher

einsteinx2:
## Summary
There is an extremely rare case where a specific string could cause an overflow error when calling `MPIHasher.hashString()` due to not using the overflow addition operator. I've changed the code from a regular addition to overflow addition which matches the original Objective-C logic and was the intended behavior.

 ## Testing Plan
Due to the nature of this hash function I was unable to write a unit test to reproduce the bug, but I reproduced it manually by setting `var hash: Int32 = Int32.max` to confirm that it was hypothetically possible and that switching to the overflow addition operator fixes the issue.

Screenshot 2024-01-16 at 2 09 06 PM

 - Closes https://go.mparticle.com/work/SQDSDKS-6076

### DIFF
--- a/mParticle-Apple-SDK/Utils/MPIHasher.swift
+++ b/mParticle-Apple-SDK/Utils/MPIHasher.swift
@@ -30,7 +30,7 @@ import Foundation
        
         var hash: Int32 = 0
         for byte in dataToHash {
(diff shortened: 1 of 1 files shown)


BrandonStalnaker (approved):
LGTM

PR #247, opened 2024-01-17
fix: Prevent Database Growth When ATT Status Denied

BrandonStalnaker:
## Summary
 - This change ensures that we never grow the database when ATT Status has been set to denied. Any Messages or Uploads already saved will remain until uploaded or [service provides an invalid code](https://github.com/mParticle/mparticle-apple-sdk/blob/efb81cd0241225dfbbfdeb7a262464adeea38cf8/mParticle-Apple-SDK/Network/MPNetworkCommunication.m#L579) 

 ## Testing Plan
 - Tested locally to confirm Denied ATT status blocks saving to the DB

 - Closes https://go.mparticle.com/work/SQDSDKS-5654

### DIFF
--- a/mParticle-Apple-SDK/Network/MPNetworkCommunication.m
+++ b/mParticle-Apple-SDK/Network/MPNetworkCommunication.m
@@ -394,7 +394,9 @@ static NSObject<MPConnectorFactoryProtocol> *factory = nil;
 }
 
 - (void)checkResponseCodeToDisableEventLogging:(NSInteger)responseCode {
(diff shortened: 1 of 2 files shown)


BrandonStalnaker:
@einsteinx2 So the only issue I could find should identity.mparticle.com or nativesdks.mparticle.com be blocked is that the database could grow excessively large depending on what HTTP response a blocked URL returns. This change prevents that issue by blocking events from being saved to the DB. Currently, I'm only blocking save to the database when ATT Status is denied so that users not utilizing ATT Status won't see any change.

Personally I'm worried this change might be excessively cautious since events are already blocked at the Mparticle server level based on ATT status but it is what the ticket asked for and we can't be sure what HTTP response apple will give when blocking urls.

PR #248, opened 2024-02-02
feat: Setup Logic for Tracking URL

BrandonStalnaker:
## Summary
 - Add logic to route requests to the `tracking-` urls after a user accepts tracking (using ATT status), before accepting it should route to the current urls

 ## Testing Plan
 - Tested normal function in sample app (using default URL) and updated unit tests
 - Closes https://go.mparticle.com/work/SQDSDKS-6060


einsteinx2 (approved):
Looks good, I don't see any issues. Thanks for updating the unit tests!

PR #253, opened 2024-02-22
ci: Update carthage.sh for Xcode 15 and ARM Macs

einsteinx2:
## Summary
 Our carthage.sh script used by the SDK release action did not support Xcode 15 as it requires new environment variables. Also it didn't support ARM build hosts which also need their own environment variables to get regular framework builds to use x86_64 architecture instead of arm64 like a device. This change adds the correct variables to fix both of those issues, allowing us to up our CI builds to use Xcode 15 and support the new ARM Mac Github runners that are starting to roll out. 

 ## Testing Plan
 - Build script was tested with Xcode 15 on both an M2 and Intel Mac and works correctly.

 - Closes https://go.mparticle.com/work/SQDSDKS-6168


BrandonStalnaker (approved):
LGTM

PR #261, opened 2024-03-13
fix: Correct threading around user notification log

BrandonStalnaker:
## Instructions

 ## Summary
 - Fix occasion timing crash around log notification

1.  App goes to background for a long time (not killed)
2. User receives and taps notification
3. App opens and at the same, concurrently, time attempts to
 a. on main queue, log user notification with _session property in https://github.com/mParticle/mparticle-apple-sdk/blob/d604e598b74503ff6f675c0989a014a0f266f21d/mParticle-Apple-SDK/MPBackendController.m#L1986 
 b. on internal message queue, end old session and begin a new one in https://github.com/mParticle/mparticle-apple-sdk/blob/d604e598b74503ff6f675c0989a014a0f266f21d/mParticle-Apple-SDK/MPBackendController.m#L2200 

The read on main queue in logUserNotification: method is done without synchronisation, resulting in occasional crashes on notification interactions.

 ## Testing Plan
 - Tested using sample app while manually triggering logNotification

 - Closes https://go.mparticle.com/work/SQDSDKS-6192

### DIFF
--- a/mParticle-Apple-SDK/MPBackendController.m
+++ b/mParticle-Apple-SDK/MPBackendController.m
@@ -1960,39 +1960,41 @@ static BOOL skipNextUpload = NO;
 }
 
 - (void)logUserNotification:(MParticleUserNotification *)userNotification {
-    NSMutableDictionary *messageInfo = [@{kMPPushNotificationStateKey:userNotification.state,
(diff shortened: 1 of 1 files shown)


einsteinx2 (approved):
Looks like a solid solution to the issue.

PR #263, opened 2024-03-14
feat: Improve mParticle reset methods

einsteinx2:
## Summary
 This looks big because of all the test cleanup, but the only real changes are in the `mParticle.m/h` files and are minimal
 
 - Revert old `reset` method to it's original implementation and mark it deprecated
 - Update new `reset:` method to use async dispatch to prevent potential for deadlocks
 - Clean up tests and ensure use of new reset method

 ## Testing Plan
 - All tests pass and performed E2E test to confirm workspace switching still works as expected

 - Closes https://go.mparticle.com/work/SQDSDKS-6229

### DIFF
--- a/UnitTests/BracketTests.mm
+++ b/UnitTests/BracketTests.mm
@@ -8,14 +8,6 @@
 
 @implementation BracketTests
 
-- (void)setUp {
-    [super setUp];
-}
-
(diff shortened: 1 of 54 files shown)


BrandonStalnaker (approved):
LGTM

PR #264, opened 2024-03-15
ci: Increase timeout on unit test expectations

einsteinx2:
## Summary
 After doing some digging it appears that the errors and other failed tests related to the kit registry were a red herring. What it appears was happening was that on a slow CI runner if one of the tests with chained calls to dispatch_after would time out, it would move on to the next tests but then a few seconds later it would finish which interfered with the data of other tests. The timing was set pretty tight, or rather plenty of time on a fast MacBook Pro, but tight on a slow CI runner. I've increased all timeouts to 15 seconds, which should be more than enough time for all runners to complete.

 ## Testing Plan
 - Tests ran fine locally, but they always did... I'll be running them multiple times via this PR to confirm it succeeds every time.

 - Closes https://go.mparticle.com/work/SQDSDKS-6230

### DIFF
--- a/UnitTests/MPBackendControllerTests.m
+++ b/UnitTests/MPBackendControllerTests.m
@@ -24,8 +24,6 @@
 #import <CoreLocation/CoreLocation.h>
 #endif
 
-#define BACKEND_TESTS_EXPECTATIONS_TIMEOUT 10
-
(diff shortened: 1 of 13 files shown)


BrandonStalnaker (approved):
LGTM

PR #267, opened 2024-03-25
feat: Update Audience Logic

BrandonStalnaker:
## Summary
 - Add methods to the public API to retrieve audience IDs associated with the current user

 ## Testing Plan
 - Tested locally with hard coded data.

 - Closes https://go.mparticle.com/work/SQDSDKS-6150


einsteinx2 (approved):
This looks ok for now, I'll do a more in depth review and comparison with the Web and Android SDK implementations when we PR this into development.

PR #270, opened 2024-04-16
fix: Type hinting MPIdentityApiRequest.identities

einsteinx2:
## Summary
 Fix crash when accessing the `-[MPIdentityApiRequest identity]` property from Swift due to incorrect type hinting

 We had it hinted as `<NSNumber*, NSString*>` but then we also stored `NSNull` objects in some cases. That was fine with Objective-C since it ignores the type hints at runtime, but Swift doesn't and crashes if any NSNull objects are present when accessing the `identities` property.

 Note this doesn't actually change the underlying behavior of this property in any way, it just now is correctly type hinted as `<NSNumber*, NSObject*>` to include the possibility of `NSNull`.

 ## Testing Plan
 - First reproduced the crash, then confirmed that I can now access the property from Swift without crashing. Also added unit tests to confirm both the core behavior and the Swift interop, explicitly accessing the property from Swift when it contains an NSNull object.

 - Closes https://go.mparticle.com/work/SQDSDKS-6338

### DIFF
--- a/UnitTests/MPIdentityApiRequestTests.m
+++ b/UnitTests/MPIdentityApiRequestTests.m
@@ -6,6 +6,10 @@
 
 @end
 
+@interface MPIdentityApiRequest ()
+@property (nonatomic) NSMutableDictionary<NSNumber*, NSObject*> *mutableIdentities;
+@end
+
 @implementation MPIdentityApiRequestTests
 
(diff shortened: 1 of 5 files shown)


BrandonStalnaker (approved):
LGTM

PR #275, opened 2024-04-30
fix: Allow Alias When Start and End Time Not Set

BrandonStalnaker:
## Summary
 - Changed functionality to handle situations when start and end time is nil more [like Android does](https://github.com/mParticle/mparticle-android-sdk/blob/2698bde6463d034258382b14e6fe5b824b56c9e4/android-core/src/main/java/com/mparticle/identity/AliasRequest.java#L159). Including using default start and end times if not set by the user.

 ## Testing Plan
 - Tested manually on sample app and updated unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6324


einsteinx2 (approved):
Discussed on Zoom, LGTM

PR #276, opened 2024-04-30
fix: Forwarding ATT status to kits

einsteinx2:
## Summary
 - We already had an optional protocol method for this but weren't actually forwarding when set on mParticle object. This adds that forwarding as it's necessary for the latest Braze SDK update and iOS 17 app tracking stuff.

 ## Testing Plan
 - E2E tested to see that the kit correctly received the status, tested with both an NSNumber and nil for the timestamp.

 - Closes https://go.mparticle.com/work/SQDSDKS-6409


BrandonStalnaker (approved):
Walked through the changes with Ben on a call. LGTM!

PR #277, opened 2024-05-10
fix: Remove unnecessary badge number functionality

einsteinx2:
## Summary
 We received a report of freezing causing due to retrieving the badge number count. While this appears to be an Apple bug, we not longer have any need for that information, so we've removed it entirely to prevent the potential freeze.

 ## Testing Plan
 - Performed SDK smoke test using a test app

 - Closes https://go.mparticle.com/work/SQDSDKS-6416


BrandonStalnaker (approved):
LGTM

PR #279, opened 2024-06-07
fix: Remove deprecated iAd framework

einsteinx2:
## Summary
 - Remove deprecated and no longer functional iAd framework code
 - Replace unnecessary reflection with direct calls to AdServices framework (this is already guarded by the `if (@available(14.3` check, so we can just call the method directly
 - Fixed incorrect passing of error pointer instead of memory address which had been obfuscated by the use of reflection

 ## Testing Plan
 - Smoke test in iOS and tvOS test app

 - Closes https://go.mparticle.com/work/SQDSDKS-6525

### DIFF
--- a/mParticle-Apple-SDK/Utils/MPSearchAdsAttribution.m
+++ b/mParticle-Apple-SDK/Utils/MPSearchAdsAttribution.m
@@ -5,7 +5,6 @@
 #import "MPILogger.h"
 
 #if TARGET_OS_IOS == 1
-#import <iAd/ADClient.h>
 #import <AdServices/AAAttribution.h>
 #endif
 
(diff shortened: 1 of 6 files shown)


mParticle-Apple-SDK/Utils/MPSearchAdsAttribution.m
```diff
@@ -18,27 +17,15 @@ + (dispatch_queue_t)messageQueue;
 
 @implementation MPSearchAdsAttribution 
 
+
```
BrandonStalnaker, line 20: extra line?

BrandonStalnaker (approved):
LGTM just a minor nitpick

mParticle-Apple-SDK/Utils/MPSearchAdsAttribution.m
einsteinx2: Oh weird, I'll remove it.

einsteinx2, line 20:
```suggestion
```

PR #280, opened 2024-06-07
fix: Record PushRegistration Correctly in UI

BrandonStalnaker:
## Summary
 - The push registration notification was not passing the correct userinfo. This made the UI read it as an unregister rather than a register

 ## Testing Plan
 - Tested in a sample app and confirmed end to end

 - Closes https://go.mparticle.com/work/SQDSDKS-6516


einsteinx2 (approved):
Love simple fixes :)

approved

PR #286, opened 2024-08-12
feat: Cert Pinning Update

BrandonStalnaker:
## Instructions

 ## Summary
 - Check only the root cert and update allowed certs

 ## Testing Plan
 - Tested in sample 

 - Closes https://go.mparticle.com/work/SQDSDKS-6545


einsteinx2 (approved):
reviewed live

PR #288, opened 2024-09-04
fix: Properly attribute events when workspace switches on new launch

einsteinx2:
## Summary
 - While working on the previous workspace switching feature, I found this edge case where launching the app with a new workspace (instead of calling switchWorkspace while app is running) will result in any waiting, un-batched messages  to be uploaded to the new workspace instead of the previous one. 

 ## Testing Plan
 - Tested with 2 workspaces and confirmed un-batched events from previous workspace upload to the previous workspace as intended. 

 - Closes https://go.mparticle.com/work/SQDSDKS-6648

### DIFF
--- a/mParticle-Apple-SDK/Data Model/MPUpload.h
+++ b/mParticle-Apple-SDK/Data Model/MPUpload.h
@@ -3,6 +3,7 @@
 #import "MPEnums.h"
 
 @class MPSession;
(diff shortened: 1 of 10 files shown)


BrandonStalnaker (approved):
LGTM Talked over on Zoom a couple times for review

PR #290, opened 2024-09-13
fix: Prevent Repeated Sessions on active Background

BrandonStalnaker:
## Summary
 - Do not automatically start a new session after we close out an idle background session.

 ## Testing Plan
 - Tested in sample app using music player to test activity in background

 - Closes https://go.mparticle.com/work/SQDSDKS-6724


einsteinx2 (approved):
Reviewed over Zoom

PR #291, opened 2024-09-16
refactor: Port MPLocationManager to Swift

einsteinx2:
## Summary
 - Port MPLocationManager to Swift

 ## Testing Plan
 
Extensively tested in a sample app (both "always" and "when in use" options, and confirmed receiving delegate callbacks)

 - Closes https://go.mparticle.com/work/SQDSDKS-6730


BrandonStalnaker (approved):
Reveiwed over Slack. LGTM

PR #295, opened 2024-09-25
refactor: Port MPILogger to Swift

BrandonStalnaker:
## Summary
 - Port MPILogger to Swift. Do not replace existing implementation as Swift version can only be used in Swift.

 ## Testing Plan
 - Tested on sample app in simulator. Confirmed all unit tests as well

 - Closes https://go.mparticle.com/work/SQDSDKS-6476


mParticle-Apple-SDK/Logger/MPLogger.swift
```diff
@@ -0,0 +1,38 @@
...
+
+public struct MPLogger {
+    
+    public static func MPLogger(loggerLevel: MPILogLevel, format: String, arguments: any CVarArg...) {
+        if (MParticle.sharedInstance().logLevel.rawValue >= loggerLevel.rawValue && loggerLevel != .none) {
+            let msg = String.localizedStringWithFormat(format, arguments)
+// Custom Logger is marked as nonnull despite the fact that we expect it to be null if not set by the client
+//         if let customLogger = MParticle.sharedInstance().customLogger {
+//             customLogger(msg)
+//         } else
```
einsteinx2, line 18:
```suggestion
```
I assume this was meant to be deleted?

mParticle-Apple-SDK/Logger/MPLogger.swift
```diff
@@ -0,0 +1,38 @@
...
+//  mParticle-Apple-SDK
+//
+//  Created by Brandon Stalnaker on 9/22/24.
+//
+
+import Foundation
+
+public struct MPLogger {
+    
+    public static func MPLogger(loggerLevel: MPILogLevel, format: String, arguments: any CVarArg...) {
```
einsteinx2, line 12: Just a suggestion, but maybe we should mark this private to force the use of the other methods below. There should never be a need to call this one directly

einsteinx2 (changes requested):

mParticle-Apple-SDK/Logger/MPLogger.swift
BrandonStalnaker: that makes sense to me

mParticle-Apple-SDK/Logger/MPLogger.swift
BrandonStalnaker: no I wanted to leave this here to make sure we understood this is kinda a todo once we fully convert to Swift

BrandonStalnaker:
Addressed comments ready for a new look @einsteinx2

einsteinx2 (approved):
Changed nullability in public interface per my request. LGTM

PR #299, opened 2024-10-10
fix: Potential crash in MPStateMachine when entering foreground

einsteinx2:
## Summary
Fix for potential crash in `MPStateMachine` when entering the foreground due to potential null pointer dereference. 

 ## Testing Plan

Smoke tested E2E using test app including backgrounding/foregrounding

I also confirmed via logging pointers that the same object/pointer/address is passed into the `addObserver:` calls. 

 - Closes https://go.mparticle.com/work/SQDSDKS-6853


BrandonStalnaker (approved):
After some questions with Ben LGTM

PR #301, opened 2024-10-17
fix: forward original event names to UI for kits with mapped events

mmustafa-tse:
## Summary
 - When a kit have custom mappings (such as AppsFlyer and GA4 firebase) we currently forward the mapped event name to the UI in livestream and event forwarding view where only the outbound is increased while the inbound is 0, while for the original event the inbound is increased and the output will not. This fix will forward the reporting issue by only forwarding the reporting of the original event in the outbound and will not create a new event to the schema and only increase the outbound.

 ## Testing Plan
 - [Y] Was this tested locally? If not, explain why.
 - partial-E2E testing with breakpoints in the AppsFlyer kit to see if the mapped event forwarded as expected as well as full E2E testing on the UI to see if the original event is reported as is.
 - Unit testing cases added to check if projectedEvents are not nil and the originalEventCopy variable in kitFiler have the original event name.

 - Closes https://mparticle-eng.atlassian.net/browse/SQDSDKS-6752

### DIFF
--- a/mParticle-Apple-SDK/Data Model/MPForwardRecord.m
+++ b/mParticle-Apple-SDK/Data Model/MPForwardRecord.m
@@ -112,10 +112,13 @@ NSString *const kMPFROptOutState = @"s";
     if (kitFilter.appliedProjections.count > 0) {
         NSMutableArray *projections = [[NSMutableArray alloc] initWithCapacity:kitFilter.appliedProjections.count];
         NSMutableDictionary *projectionDictionary;
-        NSString *currentProjectionName = ((MPEvent *)kitFilter.originalEvent).name;
+        NSString *currentProjectionName;
+        if ([kitFilter.originalEvent isKindOfClass:[MPEvent class]]) {
+            currentProjectionName = ((MPEvent *)kitFilter.originalEvent).name;
+        }
         
         for (MPEventProjection *eventProjection in kitFilter.appliedProjections) {
-            if (eventProjection.projectedName == currentProjectionName) {
+            if ([eventProjection.projectedName isEqual:currentProjectionName]) {
                 projectionDictionary = [[NSMutableDictionary alloc] initWithCapacity:4];
                 projectionDictionary[kMPFRProjectionId] = @(eventProjection.projectionId);
                 projectionDictionary[kMPMessageTypeKey] = NSStringFromMessageType(messageType);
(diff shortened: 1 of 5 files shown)


mParticle-Apple-SDK/Data Model/MPForwardRecord.m
```diff
@@ -112,19 +112,22 @@ - (instancetype)initWithMessageType:(MPMessageType)messageType execStatus:(MPKit
...
         for (MPEventProjection *eventProjection in kitFilter.appliedProjections) {
-            projectionDictionary = [[NSMutableDictionary alloc] initWithCapacity:4];
-            projectionDictionary[kMPFRProjectionId] = @(eventProjection.projectionId);
-            projectionDictionary[kMPMessageTypeKey] = NSStringFromMessageType(messageType);
-            
-            projectionDictionary[kMPEventTypeKey] = NSStringFromEventType(eventProjection.eventType);
-            
-            if (eventProjection.projectedName) {
-                projectionDictionary[kMPFRProjectionName] = eventProjection.projectedName;
+            if (eventProjection.projectedName == currentProjectionName) {
```
einsteinx2, line 118: In Objective-C, you can't use `==` to compare objects (including strings). The `==` will just compare if the pointers are equal, there's no operator overloading. So you need to compare using `isEqual:`
```suggestion
            if ([eventProjection.projectedName isEqual:currentProjectionName]) {
```

mParticle-Apple-SDK/Data Model/MPForwardRecord.m
```diff
@@ -112,19 +112,22 @@ - (instancetype)initWithMessageType:(MPMessageType)messageType execStatus:(MPKit
     if (kitFilter.appliedProjections.count > 0) {
         NSMutableArray *projections = [[NSMutableArray alloc] initWithCapacity:kitFilter.appliedProjections.count];
         NSMutableDictionary *projectionDictionary;
+        NSString *currentProjectionName = ((MPEvent *)kitFilter.originalEvent).name;
```
einsteinx2, line 115: You should check that `originalEvent` is an `MPEvent` before casting it.

einsteinx2 (changes requested):

mParticle-Apple-SDK/Data Model/MPForwardRecord.m
mmustafa-tse: jus added this change

mParticle-Apple-SDK/Data Model/MPForwardRecord.m
mmustafa-tse: added as well

einsteinx2 (approved):
lgtm

BrandonStalnaker (approved):
LGTM

PR #308, opened 2024-10-24
refactor: Initialize Kits in Background Thread

BrandonStalnaker:
## Summary
 - InitializeKits doesn't need to be called on the main thread

 ## Testing Plan
 - Tested in simulator with several of our most used kits

 - Closes https://go.mparticle.com/work/SQDSDKS-6616


einsteinx2 (approved):
Thanks I feel a lot better about this with the dispatch to main for the kit delegate method calls

PR #312, opened 2024-11-07
fix: Address firstInstall Edge Case

BrandonStalnaker:
## Summary
 - Deal with edge case that may have caused fewer first install reports than users

 ## Testing Plan
 - Tested in local sample app. Could not recreate original issue but confirmed this change doesn't cause any new issues

 - Closes https://go.mparticle.com/work/SQDSDKS-6899


einsteinx2 (approved):
Discussed over Zoom, this seems to be a reasonable potential solution.

PR #313, opened 2024-11-19
ci: Add CI build of React Native

BrandonStalnaker:
## Summary
 - Added .github/workflows/build-secondary-platforms.yml to build RN with the core sdk associated with each PR. 
 - Customized RNExample/ios/Podfile to use the PRs associated core sdk as a development pod
 - RNExample/ios/RNExample/AppDelegate.mm updated to call into the sdk to initialize

 ## Testing Plan
 - here goes (github action)

 - Closes https://go.mparticle.com/work/SQDSDKS-6928


.github/workflows/build-secondary-platforms.yml
```diff
@@ -0,0 +1,80 @@
+name: Secondary Platforms iOS Build
+
+on: [pull_request]
+
+jobs:
+#  pod-lint:
```
einsteinx2, line 6: Please remove all the commented out stuff in this file

.gitignore
```diff
@@ -37,3 +37,66 @@ Carthage
...
+/jest/integration/build/
+
+# Test Reports
+/reports
+
+# Stack Dumps generated when programs crash (Ex. bash.exe.stackdump on Win)
+*.stackdump
+
+# ReactCommon subdir shouldn't have Xcode project
+/packages/react-native/ReactCommon/**/*.xcodeproj
```
einsteinx2, line 87: Can we just gitignore the entire `/packages` directory? I assume these packages and their exact directories may change in the future. Also is there any need to include other things from `/packages` ever?

RNExample/android/app/proguard-rules.pro
```diff
@@ -0,0 +1,10 @@
+# Add project specific ProGuard rules here.
```
einsteinx2, line 1: Is it possible to remove the Android part of this example project entirely? I know by default it creates both, but it should be possible to make an iOS only RN project.

einsteinx2 (changes requested):

einsteinx2 (approved):

PR #320, opened 2024-12-16
refactor: Port MPListenerController to Swift

BrandonStalnaker:
## Summary
 - Refactor MPListenerController and MPListenerProtocol to Swift

 ## Testing Plan
 - Tested using sample app

 - Closes https://go.mparticle.com/work/SQDSDKS-6730


mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
+//
+//  MPListenerController.swift
+//  mParticle-Apple-SDK
+//
+//  Created by Brandon Stalnaker on 12/16/24.
+//
+
+import Foundation
+
+@objc public enum MPEndpoint: Int {
```
einsteinx2, line 10: Let's move these 2 enums and the MPListenerProtocol into a dedicated `MPListenerProtocol.swift` file

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+    @objc public class func sharedInstance() -> MPListenerController {
+        return shared
+    }
+
+    /**
+     * Adds a listener to the SDK to receive any MPListenerProtocol calls from the API to that object
+     * @param sdkListener An instance of a class that implements the MPListenerProtocol
+     */
+    @objc public func addSdkListener(_ sdkListener: any MPListenerProtocol) {
+        self.sdkListeners.append(sdkListener)
```
einsteinx2, line 124:
```suggestion
        sdkListeners.append(sdkListener)
```

Per style guide, let's get rid of these explicit `self.` calls in

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    /**
+     * Removes a listener from the SDK to no longer receive any MPListenerProtocol calls from the API to that object
+     * If you don't remove the Listener we will retain a zombie reference of your object and it will never be released
+     * @param sdkListener An instance of a class that implements the MPListenerProtocol
+     */
+    @objc public func removeSdkListener(_ sdkListener: any MPListenerProtocol) {
+        self.sdkListeners = self.sdkListeners.filter { sdkListener !== $0 }
```
einsteinx2, line 133:
```suggestion
        sdkListeners = sdkListeners.filter { sdkListener !== $0 }
```

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+//  Created by Brandon Stalnaker on 12/16/24.
+//
+
+import Foundation
+
+@objc public enum MPEndpoint: Int {
+    case identityLogin = 0, identityLogout, identityIdentify, identityModify, events, config, alias
+}
+@objc public enum MPDatabaseTable: Int {
+    case attributes = 0, breadcrumbs, messages, reporting, sessions, uploads, unknown
```
einsteinx2, line 14: This style is fine for now, but let's discuss later if we want to use this style or enforce a more classic "every item on it's own line" style.

mParticle-Apple-SDK/Listener/MPListenerController.swift
einsteinx2: Another thing, this actually has different behavior than the original Obj-C code due to the use of the `!==` identity equality operator. The original `removeObject:` call in Obj-C will use the `isEqual:` method to compare objects, which is essentially equivalent to `==` in Swift, but you're using the `===` form here (or well the negative `!==` form vs `!=`) which checks for address equality, i.e. exact same object.

In this case it may not matter, but was this intentional?

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+     * @param apiName the name of the API method
+     * @param parameter1 to parameter4 are the arguments sent to this api, such as the MPEvent in logEvent
+     */
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?, parameter2: NSObject?, parameter3: NSObject?, parameter4: NSObject?) {
+        
+    }
+
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?, parameter2: NSObject?, parameter3: NSObject?) {
+        for delegate in self.sdkListeners {
+            let stackTrace = [Thread.callStackSymbols]
```
einsteinx2, line 148:
```suggestion
            let stackTrace = Thread.callStackSymbols
```

This has different behavior from the original Obj-C. The `Thread.callStackSymbols` method returns an array (`[String]`) already, so this is putting an array inside another array.

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+            }
+            DispatchQueue.main.async {
+                delegate.onAPICalled?(NSStringFromSelector(apiName), stackTrace: stackTrace, isExternal: true, objects: parameters)
+            }
+        }
+    }
+
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?, parameter2: NSObject?) {
+        for delegate in self.sdkListeners {
+            let stackTrace = [Thread.callStackSymbols]
```
einsteinx2, line 173:
```suggestion
            let stackTrace = Thread.callStackSymbols
```

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?, parameter2: NSObject?, parameter3: NSObject?, parameter4: NSObject?) {
+        
+    }
+
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?, parameter2: NSObject?, parameter3: NSObject?) {
+        for delegate in self.sdkListeners {
+            let stackTrace = [Thread.callStackSymbols]
+            var parameters: [NSObject] = []
+            if parameter1 != nil {
+                parameters.append(parameter1!)
```
einsteinx2, line 151:
```suggestion
            if let parameter1 = parameter1 {
                parameters.append(parameter1)
```

All of these nil checks should be replaced with `if let` and all uses of the `!` operator removed.

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+
+    /**
+     * Indicates that an API method was called. This includes invocations both from external sources (your code)
+     * and those which originated from within the SDK
+     * @param apiName the name of the API method
+     * @param stackTrace is the current stackTrace as an array of NSStrings
+     * @param isExternal true, if the call originated from outside of the SDK
+     * @param objects is the arguments sent to this api, such as the MPEvent in logEvent
+     */
+    @objc optional func onAPICalled(_ apiName: Any!, stackTrace: Any!, isExternal: Any!, objects: Any!)
```
einsteinx2, line 27: I assume the `!`'s here are to keep the same public interface as before?

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,353 @@
...
+     * and those which originated from within the SDK
+     * @param apiName the name of the API method
+     * @param parameter1 to parameter4 are the arguments sent to this api, such as the MPEvent in logEvent
+     */
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?, parameter2: NSObject?, parameter3: NSObject?, parameter4: NSObject?) {
+        
+    }
+
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?, parameter2: NSObject?, parameter3: NSObject?) {
+        for delegate in self.sdkListeners {
```
einsteinx2, line 147:
```suggestion
        for delegate in sdkListeners {
```

All of these explicit calls to self should be removed per the style guide

einsteinx2 (changes requested):

einsteinx2:
Also I notice this doesn't seem to be used in any tests? We should add some before merging this.

mParticle-Apple-SDK/Listener/MPListenerController.swift
BrandonStalnaker: It was intentional because an array of this type [MPListenerProtocol] doesn't have any properties to compare so I had to check for adress (basically == is not valid for comparing 2 MPListenerProtocol conforming objects in swift)

mParticle-Apple-SDK/Listener/MPListenerController.swift
BrandonStalnaker: yep line 43 of mParticle-Apple-SDK/Include/MPListenerProtocol.h

mParticle-Apple-SDK/Listener/MPListenerController.swift
```diff
@@ -0,0 +1,256 @@
...
+                delegate.onAPICalled?(NSStringFromSelector(apiName), stackTrace: stackTrace, isExternal: true, objects: parameters)
+            }
+        }
+    }
+
+    @objc public func onAPICalled(_ apiName: Selector, parameter1: NSObject?) {
+        for delegate in sdkListeners {
+            let stackTrace = Thread.callStackSymbols
+            var parameters: [NSObject] = []
+            if parameter1 != nil {
```
einsteinx2, line 98:
```suggestion
            if let parameter1 = parameter1 {
```

einsteinx2 (changes requested):

einsteinx2 (approved):

PR #321, opened 2024-12-19
refactor: Port MPLaunchInfo to Swift

BrandonStalnaker:
## Summary
 - Refactored the MPLaunchInfo class to swift

 ## Testing Plan
 - Ran in sample app and confirmed existing unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6986


mParticle-Apple-SDK/Utils/MPLaunchInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+import Foundation
+@objc public class MPLaunchInfo : NSObject {
+    @objc private var sourceApp: String?
+    @objc public private(set) var url: URL
+    @objc public private(set) var sourceApplication: String?
+    @objc public private(set) var annotation: String?
+    @objc public private(set) var options: [String : Any]?
+
+    @objc(initWithURL:sourceApplication:annotation:) public init?(url: URL?, sourceApplication: String?, annotation: Any?) {
+        guard let url else { return nil }
```
einsteinx2, line 17: Just a note for discussion later to add to the style guide, but I personally really like single line guard statements like this. I think I've been doing them elsewhere with new lines like this:

```
guard let url else { 
    return nil 
}
```

But I was only doing that because we tended to not use single line statements in other places, but I would love to discuss this and add something official to the style guide about it.

mParticle-Apple-SDK/Utils/MPLaunchInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+        
+        self.annotation = convertAnnotationToString(rawAnnotation: annotation)
+        
+        self.options = [:]
+        if self.sourceApplication != nil {
+            self.options?[UIApplication.OpenURLOptionsKey.sourceApplication.rawValue] = self.sourceApplication
+        }
+        if self.annotation != nil {
+            self.options?[UIApplication.OpenURLOptionsKey.annotation.rawValue] = self.annotation
+        }
```
einsteinx2, line 39:
```suggestion
        if let sourceApplication = self.sourceApplication {
            self.options?[UIApplication.OpenURLOptionsKey.sourceApplication.rawValue] = sourceApplication
        }
        if let annotation = self.annotation {
            self.options?[UIApplication.OpenURLOptionsKey.annotation.rawValue] = annotation
        }
```

mParticle-Apple-SDK/Utils/MPLaunchInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+    
+    private func convertAnnotationToString(rawAnnotation: Any?) -> String? {
+        guard let rawAnnotation else { return nil }
+        
+        if let annotationString = rawAnnotation as? String {
+            return annotationString
+        } else if let annotationDictionary = rawAnnotation as? [String : Any] {
+            var finalDict: [String : String] = [:]
+            for (key, value) in annotationDictionary {
+                if let valueString = self.convertAnnotationToString(rawAnnotation: value) {
```
einsteinx2, line 67:
```suggestion
                if let valueString = convertAnnotationToString(rawAnnotation: value) {
```

mParticle-Apple-SDK/Utils/MPLaunchInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+                    finalDict[key] = valueString
+                }
+            }
+            if finalDict.isEmpty { return nil } else {
+                return serializeObject(finalDict)
+            }
+        } else if let annotationArray = rawAnnotation as? [Any] {
+            var finalArray: [String] = []
+            for value in annotationArray {
+                if let valueString = self.convertAnnotationToString(rawAnnotation: value) {
```
einsteinx2, line 77:
```suggestion
                if let valueString = convertAnnotationToString(rawAnnotation: value) {
```

mParticle-Apple-SDK/Utils/MPLaunchInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+                    finalArray.append(valueString)
+                }
+            }
+            if finalArray.isEmpty { return nil } else {
+                return serializeObject(finalArray)
+            }
```
einsteinx2, line 83:
```suggestion
            if finalArray.isEmpty { 
                return nil 
            } else {
                return serializeObject(finalArray)
            }
```

I'm all for the single line guard statements, but I think should be on separate lines for readability.

einsteinx2:
```suggestion
            return finalArray.isEmpty ? nil : serializeObject(finalArray)
```

Or actually even better, why not a ternary operator (if Swift allows it) in this case?

einsteinx2 (changes requested):

mParticle-Apple-SDK/Utils/MPLaunchInfo.swift
BrandonStalnaker: tested to confirm and yeah works fine

einsteinx2 (approved):

PR #326, opened 2025-01-29
refactor: Update NSDictionary+CaseInsensitive to Swift

BrandonStalnaker:
## Summary
 - Port NSDictionary(MPCaseInsensitive) to Swift
 
 ## Testing Plan
 - ran in sample app and confirmed all unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6985


mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.swift
```diff
@@ -0,0 +1,141 @@
...
+
+    @objc func caseInsensitiveKey(_ key: String) -> String? {
+        var localKey = key
+        
+        self.allKeys.forEach { keyValue in
+            if let stringKey = keyValue as? String, stringKey.caseInsensitiveCompare(key) == .orderedSame {
+                localKey = stringKey
+            }
+        }
+        return localKey
```
einsteinx2, line 19: This has slightly different behavior than the original Obj-C code. In the original, it stops searching after finding the correct key, but in this code it will continue. It's not possible to use break here as it it's an actual for loop, and if you use return it will return from the closure, but I believe it will continue to loop. Instead this can be written as a for in loop:

```suggestion
        for aKey in allKeys {
            if let stringKey = aKey as? String, stringKey.caseInsensitiveCompare(key) == .orderedSame {
                return stringKey
            }
        }
        return key
```

mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.swift
```diff
@@ -0,0 +1,141 @@
...
+    @objc func value(forCaseInsensitiveKey key: String) -> Any? {
+        var value : Any?
+        
+        self.allKeys.forEach { keyValue in
+            if let stringKey = keyValue as? String, stringKey.caseInsensitiveCompare(key) == .orderedSame {
+                value = self[stringKey]
+            }
+        }
+        
+        return value
```
einsteinx2, line 31: Same here. Also caqn be slightly simplified since with a `for in` loop you can get the key and value directly:

```suggestion
        for (aKey, value) in self {
            if let stringKey = aKey as? String, stringKey.caseInsensitiveCompare(key) == .orderedSame {
                return value
            }
        }
        return nil
```

mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @objc func transformValuesToString() -> [String : Any] {
```
einsteinx2, line 34: This method really needs a dedicate unit test written in Objective-C that passes in a dictionary of all of the supported types from all the ifs and confirms the output strings. First make sure it passes against the original Obj-C implementation then against this one.

mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.swift
```diff
@@ -0,0 +1,141 @@
...
+        let transformedDictionary: [String : Any] = self.reduce(into: [:]) { result, element in
```
einsteinx2, line 35: Double check that the use of the reduce method here doesn't bridge the NSDictionary to a Swift dict and mess up some of the types... The added unit test should catch that though.

mParticle-Apple-SDK/Utils/NSDictionary+MPCaseInsensitive.swift
```diff
@@ -0,0 +1,141 @@
...
+            } else {
+                MPLog.warning("Data type is not supported as an attribute value: \(type(of: element.value)) for key \(key)")
+            }
+        }
+        
+        return transformedDictionary
+    }
+}
+
+extension Dictionary {
```
einsteinx2, line 76: Once the tests are fixed and new tests added, this can probably be simplified so that you only write the implementations for NSDictionary or Swift dictionary and then cast to the other type. That way you don't need to duplicate the implementations. 

For sure it can be done for the `caseInsensitiveKey` and `value(forCaseInsensitiveKey key: String)` methods, but  `transformValuesToString` may have issues due to bridging messing up the types, though it would probably still be fine if the Swift implementation cases to NSDictionary vs the other way around.

einsteinx2 (changes requested):
I see that the `MPKitContainerTests testForwardAppsFlyerEvent` test is failing on this branch due to these changes. I checked out the development branch and confirmed it passes there. 

Once that test is fixed and new dedicated tests added for these methods, I can re-review and make any final suggestions to simplify the implementation.

PR #330, opened 2025-02-24
feat: update to new audience logic

BrandonStalnaker:
## Summary
 - Recover Audience API changes

 ## Testing Plan
 - Tested locally and through unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-6279


mParticle-Apple-SDK/Network/MPNetworkCommunication.m
```diff
@@ -209,6 +219,34 @@ - (MPURL *)eventURLForUpload:(MPUpload *)mpUpload {
     return eventURL;
 }
 
+- (MPURL *)audienceURL {
+    MPStateMachine *stateMachine = [MParticle sharedInstance].stateMachine;
+    
+    NSString *eventHost = [MParticle sharedInstance].networkOptions.eventsHost ?: self.defaultEventHost;
+    NSString *audienceURLFormat = [audienceFormat stringByAppendingString:@"?mpid=%@"];
+    NSString *urlString = [NSString stringWithFormat:audienceURLFormat, kMPURLScheme, self.defaultEventHost, kMPAudienceVersion, stateMachine.apiKey, kMPAudienceURL, [MPPersistenceController mpId]];
```
rmi22186, line 227: should this be `eventHost` and not `self.defaultEventHost`?
```suggestion
    NSString *urlString = [NSString stringWithFormat:audienceURLFormat, kMPURLScheme, eventHost, kMPAudienceVersion, stateMachine.apiKey, kMPAudienceURL, [MPPersistenceController mpId]];
```

rmi22186, line 227: this line is probably not needed since it's being set in 232/235

rmi22186 (changes requested):

mParticle-Apple-SDK/Network/MPURLRequestBuilder.m
```diff
@@ -187,33 +203,27 @@ - (NSMutableURLRequest *)build {
...
+            
+            MPIUserDefaults *userDefaults = [MPIUserDefaults standardUserDefaults];
+            NSString *eTag = userDefaults[kMPHTTPETagHeaderKey];
+            NSDictionary *config = [userDefaults getConfiguration];
+            if (eTag && config) {
+                [urlRequest setValue:eTag forHTTPHeaderField:@"If-None-Match"];
             }
+            
+            NSString *query = [_url.defaultURL query];
+            signatureMessage = [NSString stringWithFormat:@"%@\n%@\n%@?%@", _httpMethod, date, relativePath, query];
```
rmi22186, line 224: should this be in this PR?  i'm unclear as to what is going on here.  zoom might be easier.

BrandonStalnaker: So when I was testing this existing code with the endpoint it wasn't working because of a url with capitalization issues and because the the signature was malformed. There was also several headers being sent in the audience call that weren't necessary.  So since this was the only ticket I had for this functionality, I fixed it all in this ticket though some of this is also from the changes in that feature branch.

mParticle-Apple-SDK/Network/MPNetworkCommunication.m
BrandonStalnaker: OK so now that I'm looking at this myself I notice that on line 228 we are using this to create the default URL which is then used to create the MPURL on line 244 (which has a modified and default url within it) so it is needed

rmi22186: thanks for pointing that out

rmi22186 (approved):
LGTM.

rmi22186 (approved):

PR #336, opened 2025-03-18
fix: Issue when setting Identity Attributes to nil

BrandonStalnaker:
## Summary
 - Setting identity to nil in an identity request was causing a crash once it was reaching swift code

 ## Testing Plan
 - Tested locally and through unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-7144


rmi22186 (approved):
LGTM - Hopped on with Brandon to E2E test using old code to repro the issue, then using the new code, the sdk does not crash because Swift gets a nil value instead of NSNull

PR #338, opened 2025-03-19
refactor: Standardize Upload Data and Correct Timezone Offset Issue

BrandonStalnaker:
## Summary
 - Confirmed identical upload for upload with same events

 ## Testing Plan
Compared upload data manually, confirmed successful upload (it was before this change), and confirmed data appeared in the live stream (was not before this)


rmi22186 (approved):
lgtm

PR #339, opened 2025-03-21
feat: Add Rokt First Party Kit Support

BrandonStalnaker:
## Summary
 - Added MPRokt class with `executeWithViewName` method that call down into the Rokt sdk through the Rokt kit integrations

 ## Testing Plan
 - [ ] Was this tested locally? If not, explain why.
 - Still in the process of testing (need the changes to both core and the kit integration to full test)

 - Closes https://go.mparticle.com/work/SQDSDKS-7156


mParticle-Apple-SDK/Include/MPKitProtocol.h
```diff
@@ -132,6 +132,16 @@
 - (nullable NSString *)surveyURLWithUserAttributes:(nonnull NSDictionary *)userAttributes;
 - (BOOL) shouldDelayMParticleUpload;
 - (nonnull NSArray<MPForwardRecord *> *)logBatch:(nonnull NSDictionary *)batch;
+
+#pragma mark First Party Kits
+- (nonnull MPKitExecStatus *)executeWithViewName:(NSString * _Nullable)viewName
+                                      attributes:(NSDictionary<NSString *, NSString *> * _Nonnull)attributes
+                                      placements:(NSDictionary<NSString *, id> * _Nullable)placements
+                                          onLoad:(void (^ _Nullable)(void))onLoad onUnLoad:(void (^ _Nullable)(void))onUnLoad
```
rmi22186, line 140: onUnLoad should be on the next line

mParticle-Apple-SDK/Include/mParticle.h
```diff
@@ -85,6 +86,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+@interface MPRokt : NSObject
+
+- (void)selectPlacements:(NSString * _Nullable)identifier attributes:(NSDictionary<NSString *, NSString *> * _Nullable)attributes;
+
+- (void)selectPlacements:(NSString * _Nullable)identifier attributes:(NSDictionary<NSString *, NSString *> * _Nullable)attributes placements:(NSDictionary<NSString *, id> * _Nullable)placements onLoad:(void (^ _Nullable)(void))onLoad onUnLoad:(void (^ _Nullable)(void))onUnLoad onShouldShowLoadingIndicator:(void (^ _Nullable)(void))onShouldShowLoadingIndicator onShouldHideLoadingIndicator:(void (^ _Nullable)(void))onShouldHideLoadingIndicator onEmbeddedSizeChange:(void (^ _Nullable)(NSString * _Nonnull, CGFloat))onEmbeddedSizeChange;
```
rmi22186, line 93: can you colon align this?

mParticle-Apple-SDK/Kits/MPKitContainer.mm
```diff
@@ -2288,6 +2288,12 @@ - (void)attemptToLogEventToKit:(id<MPExtensionKitProtocol>)kitRegister kitFilter
                         return;
                     }
                     execStatus = [kitRegister.wrapperInstance logEvent:((MPEvent *)kitFilter.forwardEvent)];
+                } else if (selector == @selector(executeWithViewName:attributes:placements:onLoad:onUnLoad:onShouldShowLoadingIndicator:onShouldHideLoadingIndicator:onEmbeddedSizeChange:filteredUser:)) {
+                    if (kitFilter.shouldFilter) {
+                        return;
+                    }
+                    FilteredMParticleUser *filteredUser = [[FilteredMParticleUser alloc] initWithMParticleUser:[[[MParticle sharedInstance] identity] currentUser] kitConfiguration:self.kitConfigurations[kitRegister.code]];
+                    execStatus = [kitRegister.wrapperInstance executeWithViewName:parameters[0] attributes:parameters[1] placements:parameters[2] onLoad:parameters[3] onUnLoad:parameters[4] onShouldShowLoadingIndicator:parameters[5] onShouldHideLoadingIndicator:parameters[6] onEmbeddedSizeChange:parameters[7] filteredUser:filteredUser];
```
rmi22186, line 2296: can we give each parameter its own line for readability?

mParticle-Apple-SDK/mParticle.m
```diff
@@ -159,6 +159,89 @@ - (NSString *)description {
 @implementation MPDataPlanOptions
 @end
 
+@implementation MPRokt
+
+- (void)selectPlacements:(NSString * _Nullable)identifier
```
rmi22186, line 164: if this is required, should it be nullable?

mParticle-Apple-SDK/mParticle.m
```diff
@@ -159,6 +159,89 @@ - (NSString *)description {
...
+              attributes:(NSDictionary<NSString *, NSString *> * _Nullable)attributes {
```
rmi22186, line 165: ditto to above comment

mParticle-Apple-SDK/mParticle.m
```diff
@@ -159,6 +159,89 @@ - (NSString *)description {
...
+    NSDictionary<NSString *, NSString *> *mappedAttributes = [MPRokt mapUserAttributes:attributes mappingDictionary:@{}];
+    for (NSString *key in mappedAttributes) {
+        [[MParticle sharedInstance].identity.currentUser setUserAttribute:key value:mappedAttributes[key]];
+    }
```
rmi22186, line 169: could you actually remove this from this implementation and wait for it to be its own feature?  from my understanding, we can't test it end to end yet, so we might as well try to get the PR into development first

mParticle-Apple-SDK/mParticle.m
```diff
@@ -159,6 +159,89 @@ - (NSString *)description {
...
+        [queueParameters addParameter:identifier];
+        [queueParameters addParameter:nil];
+        [queueParameters addParameter:nil];
+        [queueParameters addParameter:nil];
+        [queueParameters addParameter:nil];
+        [queueParameters addParameter:nil];
+        [queueParameters addParameter:nil];
+        [queueParameters addParameter:nil];
+        
+        SEL roktSelector = @selector(executeWithViewName:attributes:placements:onLoad:onUnLoad:onShouldShowLoadingIndicator:onShouldHideLoadingIndicator:onEmbeddedSizeChange:filteredUser:);
```
rmi22186, line 183: maybe you can define this `executeWithViewName:attributes:placements:onLoad:onUnLoad:onShouldShowLoadingIndicator:onShouldHideLoadingIndicator:onEmbeddedSizeChange:filteredUser` somewhere since it's being used multiple times in the core sdk, as well as the kit?

mParticle-Apple-SDK/mParticle.m
```diff
@@ -159,6 +159,89 @@ - (NSString *)description {
...
+    NSDictionary<NSString *, NSString *> *mappedAttributes = [MPRokt mapUserAttributes:attributes mappingDictionary:@{}];
+    for (NSString *key in mappedAttributes) {
+        [[MParticle sharedInstance].identity.currentUser setUserAttribute:key value:mappedAttributes[key]];
+    }
+    
+    dispatch_async(dispatch_get_main_queue(), ^{
+        // Forwarding call to kits
+        MPForwardQueueParameters *queueParameters = [[MPForwardQueueParameters alloc] init];
+        [queueParameters addParameter:identifier];
+        [queueParameters addParameter:nil];
```
rmi22186, line 175: attributes should be required as well and not nil here

mParticle-Apple-SDK/mParticle.m
rmi22186: also, what would happen if they added a `placements`?  would that get set to nil?  all these parameters are optional besides identifier and attributes, and so they should be placed also. IMO we should code this so that a customer _can_ pass in arguments, but not necessarily all of them

mParticle-Apple-SDK/mParticle.m
```diff
@@ -159,6 +159,89 @@ - (NSString *)description {
...
+        [[MParticle sharedInstance].kitContainer_PRIVATE forwardSDKCall:roktSelector
+                                                                  event:nil
+                                                             parameters:queueParameters
+                                                            messageType:MPMessageTypeEvent
+                                                               userInfo:nil
+        ];
+    });
+}
+
++ (NSDictionary<NSString *, NSString *> *) mapUserAttributes:(NSDictionary<NSString *, NSString *> * _Nonnull)attributes mappingDictionary:(NSDictionary<NSString *, NSString *> *)mappingDictionary {
```
rmi22186, line 228: colon align

UnitTests/MPRoktTests.m
```diff
@@ -0,0 +1,153 @@
...
+    id mockContainer = OCMClassMock([MPKitContainer_PRIVATE class]);
+    [[[mockInstance stub] andReturn:mockContainer] kitContainer_PRIVATE];
+    [[[mockInstance stub] andReturn:mockInstance] sharedInstance];
+    
+    // Set up test parameters
+    NSString *viewName = @"testView";
+    NSDictionary *attributes = @{@"key": @"value"};
+    
+    // Set up expectations for kit container
+    OCMExpect([mockContainer forwardSDKCall:@selector(executeWithViewName:attributes:placements:onLoad:onUnLoad:onShouldShowLoadingIndicator:onShouldHideLoadingIndicator:onEmbeddedSizeChange:filteredUser:)
```
rmi22186, line 37: having this constant elsewhere would benefit tests here also

UnitTests/MPRoktTests.m
```diff
@@ -0,0 +1,153 @@
...
+                                              event:nil
+                                         parameters:[OCMArg checkWithBlock:^BOOL(MPForwardQueueParameters *params) {
+        XCTAssertEqualObjects(params[0], viewName);
+        XCTAssertNil(params[1]);
```
rmi22186, line 41: this test passes attributes as param[1].  should this assert be 
```suggestion
        XCTAssertEqualObjects(params[1], attributes);
```

UnitTests/MPRoktTests.m
```diff
@@ -0,0 +1,153 @@
...
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.1 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+        [expectation fulfill];
+    });
+    [self waitForExpectationsWithTimeout:1.0 handler:nil];
+    
+    // Verify
+    OCMVerifyAll(mockContainer);
+}
+
+- (void)testExecuteWithValidParameters {
```
rmi22186, line 68: it's no longer `Execute`, so should this also be `testSelectPlacements...`
```suggestion
- (void)testExecuteWithValidParameters {
```

and if so, should the 1st test here be something like `testSelectPlacementsWithOnlyRequiredParameters`

UnitTests/MPRoktTests.m
```diff
@@ -0,0 +1,153 @@
...
+                                       messageType:MPMessageTypeEvent
+                                          userInfo:nil]);
+    
+    // Execute method
+    [self.rokt selectPlacements:viewName
+                     attributes:attributes];
+    
+    // Wait for async operation
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Wait for async operation"];
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.1 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
rmi22186, line 59: this isn't necessary now, but one thing we found on web is that waiting a certain amount of time led to flakey tests.   is there a library that mocks time?  cursor suggested (although I have no idea if this works):

```
   XCTestExpectation *expectation = [self expectationWithDescription:@"Kit container call"];
   OCMExpect([mockContainer forwardSDKCall:...]).andDo(^(NSInvocation *invocation) {
       [expectation fulfill];
   });
```

rmi22186 (changes requested):

mParticle-Apple-SDK/mParticle.m
BrandonStalnaker: The Rokt SDK expects a nullable for this parameter so it made sense to match it. But in my experience you do need to include that so I get what you're saying. If its nullable in the Rokt SDK I would assume that means there are viable code paths where nil is passed here

mParticle-Apple-SDK/mParticle.m
BrandonStalnaker: i actually think there's a stronger case for nil here because if you're using mParticle you basically don't need to pass attributes in. Assuming you've been adding them as user attributes

UnitTests/MPRoktTests.m
BrandonStalnaker: So if you look at the MPKitProtocol.h I left the protocol method matching the Rokt SDK method because that method has all the possible parameters a client might want to pass in. That's also why I left the expanded version of the method in the Mparticle.Rokt as an additional option beyond the super simplified method we settled on.

I don't have any problem renaming the test but I want to make sure you understand wht I'm doing here/the choices I made.

mParticle-Apple-SDK/mParticle.m
BrandonStalnaker: rarely do we colon align with only 2 parameters but I do agree it looks nice so 🤷

mParticle-Apple-SDK.xcodeproj/project.pbxproj
```diff
@@ -341,7 +341,7 @@
 		53A79D2A29CE23F700E7489F /* MPAppDelegateProxy.h in Headers */ = {isa = PBXBuildFile; fileRef = 53A79B3929CDFB1F00E7489F /* MPAppDelegateProxy.h */; };
 		53A79D2C29CE23F700E7489F /* MPIConstants.h in Headers */ = {isa = PBXBuildFile; fileRef = 53A79ACF29CDFB1F00E7489F /* MPIConstants.h */; };
 		53A79D2E29CE23F700E7489F /* MPBackendController.h in Headers */ = {isa = PBXBuildFile; fileRef = 53A79AA329CDFB1E00E7489F /* MPBackendController.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		53A79D2F29CE23F700E7489F /* MPKitConfiguration.h in Headers */ = {isa = PBXBuildFile; fileRef = 53A79B5F29CDFB1F00E7489F /* MPKitConfiguration.h */; };
+		53A79D2F29CE23F700E7489F /* MPKitConfiguration.h in Headers */ = {isa = PBXBuildFile; fileRef = 53A79B5F29CDFB1F00E7489F /* MPKitConfiguration.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
BrandonStalnaker, line 344: Make this not public again

mParticle-Apple-SDK/Include/FilteredMParticleUser.h
```diff
@@ -11,6 +11,8 @@
 
 @property(readonly, strong, nonnull) NSNumber *userId;
```
BrandonStalnaker, line 13: not public

mParticle-Apple-SDK/mParticle.m
BrandonStalnaker: @danial-rokt Are their viable code paths where `viewName` can be nil and successfully retrieve an ad?

mParticle-Apple-SDK/mParticle.m
BrandonStalnaker: pass the attributes heres

rmi22186 (approved):
LGTM!  just 1 [pending comment](https://github.com/mParticle/mparticle-apple-sdk/pull/339/commits/5785e397f1e4868791218105fe2552dc07884621#r2023172325) for @danial-rokt

mParticle-Apple-SDK/mParticle.m
danial-rokt: Although there might be a path where viewName is nil, at this point, we are heavily utilising it—especially to support multiple pages.
On the Rokt Android SDK side, it is non-nullable. We've discussed this within the team, and let's match Android for consistency.

rmi22186 (approved):

PR #343, opened 2025-04-08
feat: Add user attribute mapping

BrandonStalnaker:
## Summary
 - Allows User Attribute Mapping via the Kit Container/Kit Configuration in the SDK.

 ## Testing Plan
 - Tested locally and through unit tests. Also tested out unrelated functionality and confirmed in the live stream.

 - Closes https://go.mparticle.com/work/SQDSDKS-7184

### DIFF
--- a/UnitTests/MPRoktTests.m
+++ b/UnitTests/MPRoktTests.m
@@ -234,4 +234,30 @@
     OCMVerifyAll((id)mockContainer);
 }
 
+- (void)testGetRoktPlacementAttributes {
+    id mockInstance = OCMClassMock([MParticle class]);
+    id mockContainer = OCMClassMock([MPKitContainer_PRIVATE class]);
+    NSArray *kitConfig = @[@{
+        @"AllowJavaScriptResponse": @"True",
+        @"accountId": @12345,
+        @"onboardingExpProvider": @"None",
+        @"placementAttributes": @"[{\"jsmap\":null,\"map\":\"f.name\",\"maptype\":\"UserAttributeClass.Name\",\"value\":\"firstname\"},{\"jsmap\":null,\"map\":\"zip\",\"maptype\":\"UserAttributeClass.Name\",\"value\":\"billingzipcode\"},{\"jsmap\":null,\"map\":\"l.name\",\"maptype\":\"UserAttributeClass.Name\",\"value\":\"lastname\"}]",
+        @"sandboxMode": @"True",
+        @"eau": @0,
+        @"hs": @{
+            @"pur": @{},
+            @"reg": @{}
+        },
+        @"id": @181
+    }];
+    [[[mockContainer stub] andReturn:kitConfig] originalConfig];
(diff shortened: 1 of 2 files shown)


mParticle-Apple-SDK/mParticle.m
```diff
@@ -163,30 +163,47 @@ @implementation MPRokt
...
+            NSString *mapTo = map[@"value"];
+            if (mappedAttributes[mapFrom]) {
+                NSString * value = mappedAttributes[mapFrom];
+                [mappedAttributes removeObjectForKey:mapFrom];
+                mappedAttributes[mapTo] = value;
+            }
+        }
+        for (NSString *key in mappedAttributes) {
+            [[MParticle sharedInstance].identity.currentUser setUserAttribute:key value:mappedAttributes[key]];
+        }
```
rmi22186, line 182: can this be DRYed up into a separate function so that it's called from both `selectPlacements` methods?

mParticle-Apple-SDK/mParticle.m
```diff
@@ -163,30 +163,47 @@ @implementation MPRokt
...
+            [[MParticle sharedInstance].kitContainer_PRIVATE forwardSDKCall:roktSelector
+                                                                      event:nil
+                                                                 parameters:queueParameters
+                                                                messageType:MPMessageTypeEvent
+                                                                   userInfo:nil
+            ];
+        });
+    } else {
+        MPILogVerbose(@"[MParticle.Rokt selectPlacements:attributes:] not performed due to kit not being configured");
+    }
```
rmi22186, line 206: This can also be DRYed up with the method below using all arguments and just pass either 2, or all arguments to it.

rmi22186 (changes requested):

UnitTests/MPRoktTests.m
```diff
@@ -155,4 +164,74 @@ - (void)testSelectPlacementsExpandedWithNilParameters {
     OCMVerifyAll(mockContainer);
 }
 
+- (void)testSelectPlacementsSimpleWithMapping {
```
rmi22186, line 167: This test stubs the `getRoktPlacementAttributes` call.  

Can you add a test for `getRoktPlacementAttributes` itself, using a config like the one you put in the chat.  passing in "[{\"jsmap\":null,\"map\":\"f.name\",\"maptype\":\"UserAttributeClass.Name\",\"value\":\"firstname\"},{\"jsmap\":null,\"map\":\"zip\",\"maptype\":\"UserAttributeClass.Name\",\"value\":\"billingzipcode\"},{\"jsmap\":null,\"map\":\"l.name\",\"maptype\":\"UserAttributeClass.Name\",\"value\":\"lastname\"}]"

and expecting what you have in line 168 back?

(note that i copied and pasted what you had in #temp-rokt-client but updated the mapping from `f.name` --> `firstname` instead of email

rmi22186 (approved):

PR #345, opened 2025-04-14
feat: Add MPRoktEmbeddedView Class

BrandonStalnaker:
## Summary
 - Added a MPRoktEmbeddedView class to ensure strong typing in the future

 ## Testing Plan
 - Tested with sample app and unit tests. Because the object is expected inside a dictionary and the value isn't test until with in the Kit, our best unit test for this would be in the kit integration.

 - Closes https://go.mparticle.com/work/SQDSDKS-7200

### DIFF
--- a/UnitTests/MPRoktTests.m
+++ b/UnitTests/MPRoktTests.m
@@ -85,7 +85,8 @@
     NSString *viewName = @"testView";
     NSDictionary *attributes = @{@"key": @"value"};
     NSDictionary *finalAttributes = @{@"key": @"value", @"sandbox": @"true"};
-    NSDictionary *placements = @{@"placement": @"test"};
+    MPRoktEmbeddedView *exampleView = [[MPRoktEmbeddedView alloc] initWithFrame:CGRectZero];
+    NSDictionary *placements = @{@"placement": exampleView};
     void (^onLoad)(void) = ^{};
     void (^onUnLoad)(void) = ^{};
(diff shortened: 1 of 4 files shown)


UnitTests/MPRoktTests.m
```diff
@@ -85,7 +85,8 @@ - (void)testSelectPlacementsExpandedWithValidParameters {
     NSString *viewName = @"testView";
     NSDictionary *attributes = @{@"key": @"value"};
     NSDictionary *finalAttributes = @{@"key": @"value", @"sandbox": @"true"};
-    NSDictionary *placements = @{@"placement": @"test"};
+    MPRoktEmbeddedView *exampleView = [[MPRoktEmbeddedView alloc] initWithFrame:CGRectZero];
```
sahil-rokt, line 88: In Android it is called `RoktEmbeddedView`, should we align with the name on both the platforms ? cc: @Mansi-mParticle

BrandonStalnaker: Namespacing isn't available in Objective C or Swift so [adding prefixes](https://google.github.io/styleguide/objcguide.html#prefixes) is the standard way of preventing naming collisions.

WeiYewT (approved):
looks good to me, we will need to do something similar like android with injecting the RoktEmbeddedView at the kit level.

Mansi-mParticle (approved):
LGTM

PR #348, opened 2025-04-23
feat: Implement MPRoktEventCallback on Core

BrandonStalnaker:
## Summary
 - Moved MPRokt code out into its own class and replaced individual callbacks with 

 ## Testing Plan
 - Tested using sample app and confirmed succesfull placement along with smoke testing of core.

 - Closes https://go.mparticle.com/work/SQDSDKS-7226


BrandonStalnaker:
So the main changes to this (outside of me moving MPRokt out into its own class)
1. Add a MPRoktEventCallback class that has a property for each callback offered on the Rokt SDK  selectPlacement.
2. Update MPKitProtocols selectPlacement method to take MPRoktEventCallback as a parameter
3. Update any place that was using that old method or its selector. (MPRokt, MPRoktTests, and KitContainer mostly)

mParticle-Apple-SDK/MPRokt.m
```diff
@@ -0,0 +1,132 @@
...
+    
+    // Get the kit configuration
+    NSArray<NSDictionary *> *kitConfigs = [MParticle sharedInstance].kitContainer_PRIVATE.originalConfig.copy;
+    NSDictionary *roktKitConfig;
+    for (NSDictionary *kitConfig in kitConfigs) {
+        if (kitConfig[@"id"] != nil && [kitConfig[@"id"] integerValue] == 181) {
+            roktKitConfig = kitConfig;
+        }
+    }
+
```
alexs-mparticle, line 81: I think you should exit early if `roktConfig` comes up null.

```suggestion
        if (!roktKitConfig) {
            MPILogVerbose(@"Rokt kit configuration not found");
            return nil;
        }
```

mParticle-Apple-SDK/MPRokt.m
```diff
@@ -0,0 +1,132 @@
...
+    for (NSDictionary *kitConfig in kitConfigs) {
+        if (kitConfig[@"id"] != nil && [kitConfig[@"id"] integerValue] == 181) {
+            roktKitConfig = kitConfig;
+        }
+    }
+    
+    // Get the placement attributes map
+    NSString *strAttributeMap;
+    NSData *dataAttributeMap;
+    if (roktKitConfig != nil) {
```
alexs-mparticle, line 85: If you check for `!roktKitConfig` above, you can remove this conditional here and make it a little cleaner.

alexs-mparticle (changes requested):

alexs-mparticle (approved):

PR #350, opened 2025-05-02
fix: assign correct IDFV instead of random UUID

mmustafa-tse:
## Summary
 - A customer reported that the IDFV comes in as a random UUID every time the mParticle SDK is initialized, I was able to reproduce and confirm the behavior and applied the fix below

 ## Testing Plan
 - [X] Was this tested locally? If not, explain why.
 - E2E tested

### DIFF
--- a/mParticle-Apple-SDK/Utils/MPDevice.swift
+++ b/mParticle-Apple-SDK/Utils/MPDevice.swift
@@ -217,7 +217,8 @@ import CoreTelephony
                 if let vendor = userDefaults[Device.kMPDeviceAppVendorIdKey] as? String, vendor != Device.kMPDeviceInvalidVendorId {
                     _vendorId = vendor
                 } else {
-                    _vendorId = UUID().uuidString
+                    _vendorId = UIDevice.current.identifierForVendor?.uuidString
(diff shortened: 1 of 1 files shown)


mParticle-Apple-SDK/Utils/MPDevice.swift
```diff
@@ -217,7 +217,8 @@ import CoreTelephony
                 if let vendor = userDefaults[Device.kMPDeviceAppVendorIdKey] as? String, vendor != Device.kMPDeviceInvalidVendorId {
                     _vendorId = vendor
                 } else {
-                    _vendorId = UUID().uuidString
+                    _vendorId = UIDevice.current.identifierForVendor?.uuidString
+                    userDefaults[Device.kMPDeviceAppVendorIdKey] = _vendorId
```
BrandonStalnaker, line 221: Heres the [old functionality](https://github.com/mParticle/mparticle-apple-sdk/commit/0fda3479b85db4ad5f09a0b193f14162944d140e#diff-2f76e6505603ebc6dcc124be5432a7b9f9f43838724a814e48e395ebff7f3866L297) to compare this more easily. I think this is the correct fix comparing these

BrandonStalnaker (approved):

PR #351, opened 2025-05-12
feat: Identify if Provided Email in Rokt SelectPlacements

BrandonStalnaker:
## Summary
 - After a developer calls selectPlacements with email, we should check to see if email is on the user yet.  If it isn’t, we should call identify with email (along with all other current user identities)

 ## Testing Plan
 - Tested live and with unit tests

 - Closes https://go.mparticle.com/work/SQDSDKS-7247

### DIFF
--- a/UnitTests/MPRoktTests.m
+++ b/UnitTests/MPRoktTests.m
@@ -5,6 +5,7 @@
 #import "MPIdentityApiManager.h"
 #import "MPKitContainer.h"
 #import "MPForwardQueueParameters.h"
+#import "MParticleSwift.h"
 #import "MPIConstants.h"
 
 @interface MPRokt ()
@@ -13,6 +14,11 @@
 @end
 
 @interface MPRokt (Testing)
+@end
+
+@interface MParticle ()
+@property (nonatomic, strong, readonly) MPStateMachine_PRIVATE *stateMachine;
+@property (nonatomic, strong, nonnull) MPBackendController_PRIVATE *backendController;
 @end
 
 @interface MPIdentityApi ()
@@ -295,7 +301,7 @@
     OCMVerifyAll(mockContainer);
 }
 
-- (void)testTriggeredIdentify {
+- (void)testTriggeredIdentifyWithNoIdentities {
     MParticleUser *currentUser = [MParticle sharedInstance].identity.currentUser;
 
     //Mock Identity as needed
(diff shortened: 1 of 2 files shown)


mParticle-Apple-SDK/MPRokt.m
```diff
@@ -135,4 +141,22 @@ - (void)selectPlacements:(NSString *)identifier
...
+    if (email && email != user.identities[@(MPIdentityEmail)]) {
+        MPIdentityApiRequest *identityRequest = [MPIdentityApiRequest requestWithUser:user];
+        identityRequest.email = email;
+        
+        [[[MParticle sharedInstance] identity] identify:identityRequest completion:^(MPIdentityApiResult *_Nullable apiResult, NSError *_Nullable error) {
+            if (error) {
+                NSLog(@"Failed to automatically update email from selectPlacement: %@", error);
+                completion(user);
+            } else {
+                NSLog(@"Updated email identity based off selectPlacement's attributes: %@", apiResult.user.identities.description);
```
rmi22186, line 154: Does this provide a description of all the identities that are on the user?  Cursor is telling me something like the following would come back:
{
    7 = "user@example.com";  // MPIdentityEmail
    1 = "customer123";       // MPIdentityCustomerId
    2 = "fb123456";         // MPIdentityFacebook
    4 = "google123";        // MPIdentityGoogle
    3 = "twitter123";       // MPIdentityTwitter
    5 = "ms123456";         // MPIdentityMicrosoft
    6 = "yahoo123";         // MPIdentityYahoo
    19 = "+1234567890";     // MPIdentityMobileNumber
}
```

Is that accurate? If so, i think having it be by number is confusing.

I think just confirming which email was used is fine here also.  **_Or_** if you add the above suggestion to give more detailed emails in the above log I recommended, then it would be redundant here and could be removed.

File: UnitTests/MPRoktTests.m [hunk 2]
```diff
@@ -260,4 +261,33 @@ - (void)testGetRoktPlacementAttributesMapping {
     XCTAssertEqualObjects(testResult, expectedResult, @"Mapping does not match .");
 }
 
+- (void)testSelectPlacementsIdentifyUser {
```
rmi22186, line 264: Can you add a test where there is an email (foo@gmail.com) and (bar@gmail.com) is set on `selectPlacements` - then assert that identify was called with bar@gmail.com?

mParticle-Apple-SDK/MPRokt.m
```diff
@@ -135,4 +141,22 @@ - (void)selectPlacements:(NSString *)identifier
     return finalAttributes;
 }
 
+- (void)confirmEmail:(NSString * _Nullable)email user:(MParticleUser * _Nullable)user completion:(void (^)(MParticleUser *_Nullable))completion {
+    if (email && email != user.identities[@(MPIdentityEmail)]) {
+        MPIdentityApiRequest *identityRequest = [MPIdentityApiRequest requestWithUser:user];
```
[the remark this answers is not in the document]
BrandonStalnaker: That kinda reads like an error message to me and might lead customers to think they are doing something wrong. Is that your intent? Maybe the second sentence could be more like. `We will attempt to update the current user with this email but to receive a Rokt Layout as quickly as possible, sync the email identity to mParticle as soon as you receive it.`

mParticle-Apple-SDK/MPRokt.m
BrandonStalnaker: yeah unfortunately identities are stored with the key being an integer. I could change it to just be the email.

UnitTests/MPRoktTests.m
BrandonStalnaker: I'll try. Identity is really tricky to mock well but it should be doable

mParticle-Apple-SDK/MPRokt.m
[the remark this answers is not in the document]
rmi22186: I think actually there are 2 scenarios here that should be considered for messaging purposes.  both would result in calling identify, but one i think _is_ a bug

1. if (selectPlacements email &&  !currentUser.email) // this is not a bug, as there's a chance that someone is checking out as guest, and so they'd only get the email once
2. if (selectPlacements email && currentUser.email && selectPlacements email !== currentUser.email) --> this i think is a bug.  if the user has an email, then the selectPlacements email _should_ match it.  we should log that as a warning at a minimum to the user to let them know
the first scenario doesn't need any logging since it's a valid case, the 2nd one does need a log

rmi22186 (approved):
lgtm!

PR #353, opened 2025-05-21
fix: MPAppNotificationHandler webpageURL issue

BrandonStalnaker:
## Summary
 - Prevent nil webpageURL from crashing SDK

 ## Testing Plan
 - Tested through unit tests and standard smoke tests though was not able to recreate the original crash prior to fix.

 - Closes https://go.mparticle.com/work/SQDSDKS-7334

### DIFF
--- a/UnitTests/MPAppNotificationHandlerTests.m
+++ b/UnitTests/MPAppNotificationHandlerTests.m
@@ -84,4 +84,19 @@
     [appNotificationHandler openURL:url options:options];
 }
 
+- (void)testContinueUserActivityNoCrash {
+    MPAppNotificationHandler *appNotificationHandler = [MParticle sharedInstance].appNotificationHandler;
+    
+    NSUserActivity *userActivity = [[NSUserActivity alloc] initWithActivityType:@"test"];
+    userActivity.webpageURL = [[NSURL alloc] initWithString:@"http://mparticle.com"];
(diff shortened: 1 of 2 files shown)


UnitTests/MPAppNotificationHandlerTests.m
```diff
@@ -84,4 +84,19 @@ - (void)testOpenURLOptions {
...
 
+- (void)testContinueUserActivityNoCrash {
+    MPAppNotificationHandler *appNotificationHandler = [MParticle sharedInstance].appNotificationHandler;
+    
+    NSUserActivity *userActivity = [[NSUserActivity alloc] initWithActivityType:@"test"];
+    userActivity.webpageURL = [[NSURL alloc] initWithString:@"http://mparticle.com"];
+    [appNotificationHandler continueUserActivity:userActivity restorationHandler:^(NSArray<id<UIUserActivityRestoring>> *restorationHandler){}];
+}
+
+- (void)testContinueUserActivityWithNilURLNoCrash {
```
alexs-mparticle, line 95: On the Web SDK, we sometimes do tests for invalid characters in a URL. Is there any value for testing that sort of thing here?

alexs-mparticle: No, I'm hoping that Apple tested that well enough 😆

alexs-mparticle (approved):

PR #354, opened 2025-06-02
fix: Remove alreadySynchedUserIdentities

BrandonStalnaker:
## Summary
 - On iOs we're preventing identity values from being forwarded to kits if they'd already been sent in the past. This code was very old and did not match the behavior of the Android SDK. Removing this code ensures that we favor over sending data over our clients potentially not recieving the data and brings iOs and Ansroid into better parity.

 ## Testing Plan
 - [x] Was this tested locally? Yes
 - Tested using simulator and confirmed data flowing to the Appsflyer Kit

 - Closes https://go.mparticle.com/work/SQDSDKS-7304

### DIFF
--- a/mParticle-Apple-SDK/Kits/MPKitContainer.mm
+++ b/mParticle-Apple-SDK/Kits/MPKitContainer.mm
@@ -2097,13 +2097,7 @@ static const NSInteger sideloadedKitCodeStartValue = 1000000000;
                     }
                 }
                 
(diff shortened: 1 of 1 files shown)


samdozor:
I think it's generally the right move to do this - but the reason this is here is b/c without - customers' braze bills may raise significantly. I think if we remove this - we need to add back something similar in braze to avoid huge bills for customers.

BrandonStalnaker:
@samdozor Is that not an issue for Android for some other reason? How should we move forward?

PR #361, opened 2025-06-11
feat: Multiple select placements calls made

jamesnrokt:
## Instructions

 ## Summary
 - Multiple select placements calls made due to completion block
 - This PR addresses it by only returning if user.identities does not match email
 - Tests refactored as mocks were affecting tests in other parts of code now they stop mocking after each test

 ## Testing Plan
 - [ ] Was this tested locally? If not, explain why.

 - Closes https://go.mparticle.com/work/SQDSDKS-7396


BrandonStalnaker (approved):
Great catch! LGTM

PR #365, opened 2025-06-13
chore: Add Rokt Events API to the Example App

BrandonStalnaker:
## Summary
 - Added use of the events api to the example app

 ## Testing Plan
 - tested with sample app

### DIFF
--- a/Example/mParticleExample/ViewController.m
+++ b/Example/mParticleExample/ViewController.m
@@ -187,6 +187,40 @@
                                                                @"sandbox": @"true",
                                                                @"mobile": @"(555)867-5309"
     };
+    
+    [Rokt eventsWithViewName:@"RoktLayout" onEvent:^(RoktEvent * roktEvent) {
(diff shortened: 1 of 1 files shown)


Example/mParticleExample/ViewController.m
```diff
@@ -199,6 +233,40 @@ - (void)selectDarkOverlayPlacement {
                                                                @"sandbox": @"true",
                                                                @"mobile": @"(555)867-5309"
     };
+    
+    [Rokt globalEventsOnEvent:^(RoktEvent * roktEvent) {
+        NSLog(@"Global RoktEvent triggered of type: %@", [roktEvent class]);
+
+        if ([roktEvent isKindOfClass:[InitComplete class]]) {
+            
+        } else if ([roktEvent isKindOfClass:[ShowLoadingIndicator class]]) {
```
thomson-t, line 242: Only `InitComplete` will be triggered from `globalEventsOnEvent`. Also, it is triggered when the `Rokt.init` is called. Probably we can move this to `AppDelegate`?

PR #368, opened 2025-06-17
refactor: Rename 'placements' to 'embeddedViews' in selectPlacements method and update related references

thomson-t:
## Instructions

 ## Summary
 - Rename 'placements' to 'embeddedViews' in selectPlacements method 

 ## Testing Plan
 - Tested example app and loaded overlay and embedded placements

 - Closes [SQDSDKS-7419](https://mparticle-eng.atlassian.net/browse/SQDSDKS-7419)

[SQDSDKS-7419]: https://mparticle-eng.atlassian.net/browse/SQDSDKS-7419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mParticle-Apple-SDK/Kits/MPKitContainer.mm
+++ b/mParticle-Apple-SDK/Kits/MPKitContainer.mm
@@ -2303,7 +2303,8 @@ static const NSInteger sideloadedKitCodeStartValue = 1000000000;
                     if (kitFilter.shouldFilter) {
                         return;
                     }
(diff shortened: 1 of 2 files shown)


BrandonStalnaker (approved):
LGTM but since we're changing the KitProtocol as well we should make sure the corresponding change for the kit is released at the same time.

PR #402, opened 2025-08-28
feat: SDKE-119 Create protocols for testability

denischilik:
## Summary
This is the first PR in a series focused on improving testability. Its main goal is to decouple testing code from Swift tests and establish a foundation for better testing practices.
- Introduced two new protocols and their mock implementations.
- Added a test to ensure the mocks behave as expected.
- Added a script that emits a warning when there is a duplication of the MParticle singleton within the scope of a single method.

 ## Testing Plan
 - [ x] Was this tested locally? If not, explain why.

 - Related to https://rokt.atlassian.net/browse/SDKE-119


BrandonStalnaker (approved):

BrandonStalnaker (changes requested):
Looks mostly good but your new warning is causing the linter CI to fail. 

Also one thing I'd like you to check before merging is that none of the classes that got abstracted to protocols have instances that get `isKindOfClass` called on them. Because they'd now fail. Other than that, since they're all internal classes I think we're good.

nickolas-dimitrakas (approved):
LGTM, walkthrough in person

PR #405, opened 2025-09-04
feat: SDKE-63-replace-logging-macros-with-swift-helpers

denischilik:
## Summary
 - add methods which are accessible from objc code
 - replace usage of macros in mParticle.m

 ## Testing Plan

 - Closes https://rokt.atlassian.net/browse/SDKE-63


nickolas-dimitrakas (dismissed):
LGTM

nickolas-dimitrakas (approved):

BrandonStalnaker (commented):
Waiting for unit tests to succeed but one question I have after looking this over is, is there a reason to not to delete the objective c macros in this PR?

PR #406, opened 2025-09-05
feat: SDKE-64 Improve mParticle.m test coverage in swift

denischilik:
## Summary

This PR extracts the executor logic into a dedicated Executor class implementing ExecutorProtocol. The goal is to improve testability, reduce coupling with MParticle, and centralize queue and threading logic in a single component.
It is first PR from series.

 ## Motivation

Previously, queue handling (messageQueue and main thread execution) was hard-coded inside MParticle, making the class tightly coupled and difficult to test. By moving this logic into a separate Executor class, we:

Enable easier unit testing by mocking ExecutorProtocol.

Reduce complexity and dependencies inside MParticle.

Improve maintainability and lay the groundwork for future architectural improvements.

 ## Testing Plan

 - Related to https://rokt.atlassian.net/browse/SDKE-64

### DIFF
--- a/mParticle-Apple-SDK/Executor.m
+++ b/mParticle-Apple-SDK/Executor.m
@@ -1,0 +1,62 @@
+#import "Executor.h"
+
+@implementation Executor
+
+static dispatch_queue_t messageQueue = nil;
+static void *messageQueueKey = "mparticle message queue key";
+static void *messageQueueToken = "mparticle message queue token";
+
+- (instancetype)init {
+    self = [super init];
+    if (!self) {
+        return nil;
+    }
+    messageQueue = dispatch_queue_create("com.mparticle.messageQueue", DISPATCH_QUEUE_SERIAL);
+    dispatch_queue_set_specific(messageQueue, messageQueueKey, messageQueueToken, nil);
+    
+    return self;
+}
+
+- (dispatch_queue_t)messageQueue {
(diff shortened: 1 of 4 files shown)


mParticle-Apple-SDK/Executor.m
```diff
@@ -0,0 +1,62 @@
...
+static dispatch_queue_t messageQueue = nil;
+static void *messageQueueKey = "mparticle message queue key";
+static void *messageQueueToken = "mparticle message queue token";
+
+- (instancetype)init {
+    self = [super init];
+    if (!self) {
+        return nil;
+    }
+    messageQueue = dispatch_queue_create("com.mparticle.messageQueue", DISPATCH_QUEUE_SERIAL);
```
BrandonStalnaker, line 14: If multiple Executor instances are created, won't they overwrite each other's queues? Will that cause a crash?

I still don't quite get why we are doing this hybrid pattern where
- The Executor is instantiated as an object
- But it manages static/global state

Why not just intantiate it as a strong reference on Executor?

denischilik: In the future we will create a single executor object during initialization and inject it into all dependent classes. This way there will always be exactly one shared executor instance, without relying on any static state.

nickolas-dimitrakas (approved):
LGTM

PR #411, opened 2025-09-17
Fix: SDKE-235 Crash in MPKitAPI.h

denischilik:
## Summary
 - MPAttributionResult wasn't synthesize setter because private header wasn't imported, but the core of the problem was that MPKitAPI class was using private property which wasn't available to them. We should consider to make this property readwrite in public interface instead creating private category or import related header to all places where we create this new private category.
 - create Tech Dept ticket https://rokt.atlassian.net/browse/SDKE-236
 
 ## Testing Plan

 - Closes https://rokt.atlassian.net/browse/SDKE-235


denischilik:
Correct branch https://github.com/mParticle/mparticle-apple-sdk/pull/412

PR #735, opened 2026-04-09
fix: Make setSessionId synchronous to prevent race with MPRoktLayout

nickolas-dimitrakas:
## Background

A race condition in `setSessionId` was causing web-to-native session handoff to fail on iOS. The method was delivered asynchronously via two `dispatch_async(main)` hops through `forwardSDKCall`, while `MPRoktLayout` reads the session synchronously from `UserDefaults` during execute. This meant the session ID was missing when both calls were made in quick succession. Android was unaffected because its `setSessionId` path is synchronous.

## What Has Changed

- Made `setSessionId` use the same direct kit access pattern that `getSessionId` already uses, bypassing `forwardSDKCall` and delivering the session ID synchronously to the kit
- Added fallback to `forwardSDKCall` when kits haven't initialized yet, preserving existing queuing behavior
- Replaced the old async `testSetSessionIdForwardsToKitContainer` test with three new tests covering synchronous delivery, `forwardSDKCall` bypass, and fallback behavior

## Timeline
<details>
  <summary><b>Click to see Sequence Diagram: setSessionId and MPRoktLayout Execute</b></summary>

  image

</details>

## Checklist

### DIFF
--- a/UnitTests/ObjCTests/MPRoktTests.m
+++ b/UnitTests/ObjCTests/MPRoktTests.m
@@ -11,15 +11,20 @@
 // Rokt kit identifier for testing
 static NSNumber * const kTestRoktKitId = @181;
 
-// Test helper class that simulates a kit with getSessionId method
+// Test helper class that simulates a kit with getSessionId and setSessionId methods
 @interface MPRoktTestKitInstance : NSObject
 @property (nonatomic, copy) NSString *sessionIdToReturn;
+@property (nonatomic, copy) NSString *receivedSessionId;
 - (NSString *)getSessionId;
+- (void)setSessionId:(NSString *)sessionId;
 @end
 
 @implementation MPRoktTestKitInstance
 - (NSString *)getSessionId {
     return self.sessionIdToReturn;
+}
+- (void)setSessionId:(NSString *)sessionId {
+    self.receivedSessionId = sessionId;
 }
 @end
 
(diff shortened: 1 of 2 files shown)


mParticle-Apple-SDK/MPRokt.m
```diff
@@ -309,6 +309,22 @@ - (void)close {
 ///   - sessionId: The session id to be set. Must be a non-empty string.
 - (void)setSessionId:(NSString * _Nonnull)sessionId {
     MPILogDebug(@"MPRokt setSessionId called - sessionId: %@", sessionId ? @"present" : @"nil");
+
+    NSArray<id<MPExtensionKitProtocol>> *activeKits =
+        [[MParticle sharedInstance].kitContainer_PRIVATE activeKitsRegistry];
+
+    for (id<MPExtensionKitProtocol> kitRegister in activeKits) {
```
thomson-t, line 316: `setSessionId` is only used on `rokt` object. So iterating through other kits is not required.

mParticle-Apple-SDK/MPRokt.m
```diff
@@ -309,6 +309,22 @@ - (void)close {
...
+            id kitInstance = kitRegister.wrapperInstance;
+            if (kitInstance && [kitInstance respondsToSelector:@selector(setSessionId:)]) {
+                [kitInstance performSelector:@selector(setSessionId:) withObject:sessionId];
+                MPILogDebug(@"MPRokt setSessionId - forwarded synchronously to kit");
+                return;
+            }
+        }
+    }
+
+    MPILogDebug(@"MPRokt setSessionId - kit not available, queueing for deferred execution");
```
BrandonStalnaker, line 327: So you're both setting it directly and then setting it throught the normal kit container way?

mParticle-Apple-SDK/MPRokt.m
BrandonStalnaker: since its an array he has to iterate through until he finds the rokt kit.

denischilik, line 316: We can extract loop in one method getRockKit, If we have not specific property to directly access kit

UnitTests/ObjCTests/MPRoktTests.m
```diff
@@ -11,16 +11,21 @@
 // Rokt kit identifier for testing
 static NSNumber * const kTestRoktKitId = @181;
 
-// Test helper class that simulates a kit with getSessionId method
+// Test helper class that simulates a kit with getSessionId and setSessionId methods
 @interface MPRoktTestKitInstance : NSObject
 @property (nonatomic, copy) NSString *sessionIdToReturn;
+@property (nonatomic, copy) NSString *receivedSessionId;
```
denischilik, line 17: You can just create property sessionId instead additional getter and setter.

UnitTests/ObjCTests/MPRoktTests.m
```diff
@@ -11,16 +11,21 @@
...
 - (NSString *)getSessionId;
+- (void)setSessionId:(NSString *)sessionId;
 @end
 
 @implementation MPRoktTestKitInstance
```
denischilik, line 22: In general about the test, shouldn't we create new test in swift?

mParticle-Apple-SDK/MPRokt.m
nickolas-dimitrakas: Yes, I'm still using the previous logic in the case that the kit is not configured yet

UnitTests/ObjCTests/MPRoktTests.m
nickolas-dimitrakas: Went for speed here, I will try to create the tests in swift if it is not too time consuming. Otherwise I can try to address this in a separate PR

PR #753, opened 2026-04-14
ci: cache CI dependencies to reduce PR workflow runtime

nickolas-dimitrakas:
## Background

PR workflow jobs in \`build-kits\` were taking 15-19 minutes each because third-party SDKs (Firebase, Airship, Rokt) were being downloaded and compiled from scratch on every run with no caching.

Profiled from CI run [#24418060388](https://github.com/mParticle/mparticle-apple-sdk/actions/runs/24418060388):

| Job | Duration |
|---|---|
| Pod Lint urbanairship-19 | 18m 55s |
| Build firebase-ga4-12 | 17m 49s |
| Build firebase-ga4-11 | 16m 23s |
| Build firebase-11 | 15m 59s |
| Build urbanairship-20 | 15m 13s |
| Build rokt | 13m 6s |

## What Has Changed

**build-kits.yml**
- Added `actions/cache@v5.0.5` for `~/.cocoapods` and `~/Library/Caches/CocoaPods` in `pod-lint-kits`, keyed per-kit by podspec hash + weekly window — avoids re-downloading Airship/Firebase on every lint run; weekly window ensures new upstream releases are picked up
- Added full `DerivedData` cache in `build-kits`, key computed via shell `find | shasum` across kit source files + core SDK source files (`mParticle-Apple-SDK/**`, `mParticle-Apple-SDK-Swift/**`) — enables incremental xcodebuild instead of full recompilation on warm runs; busts on any kit or core SDK change
- Added `timeout-minutes: 30` to both `pod-lint-kits` and `build-kits` jobs to enforce the target CI runtime
- Reduced `fetch-depth` from `0` → `1` (full git history not needed for builds)

**build-and-lint.yml**
- Added gem cache (`~/.gem`) for `xcodeproj`; switched to user-level gem install

**cross-platform-tests.yml**
- Added gem cache (`~/.gem`) for `cocoapods` + `cocoapods-generate`, keyed by `Podfile.lock`

**integration-tests.yml**
- Added `actions/cache@v5.0.5` for WireMock JAR, keyed by version — skips Maven Central download on cache hit

All `actions/cache` references pinned to SHA (`v5.0.5`).

## Screenshots/Video

N/A

## Checklist

- [x] Self-review completed
- [ ] Tests added or updated
- [x] Tested locally

## Additional Notes

Cold runs (first PR after a cache miss) remain as today. On warm runs (no source changes), kit builds should drop from 15-18 min to ~2-5 min via incremental xcodebuild. Monitor GitHub Actions cache storage in Settings → Actions → Caches — Firebase DerivedData is large (1-3 GB per kit) and may approach the 10 GB repo limit.


nickolas-dimitrakas:
Superseded by #981 — this PR predates the SPM-resolve watchdog/retry logic added in #867, and merging it as-is would have silently reverted that fix. #981 covers the same caching goal without touching that step, and deliberately avoids caching compiled DerivedData (the riskier part of this PR) in favor of caching only raw dependency sources.

PR #763, opened 2026-04-30
feat: Include Payment Extension by Default

BrandonStalnaker:
## Background
- Shoppable Ads relies on RoktPaymentExtension; the kit now pulls it in by default. Apps that do not want that stack can use the no-payments variant instead.

## What Has Changed
- Swift Package Manager (Kits/rokt/rokt/Package.swift)

- Declares a dependency on https://github.com/ROKT/rokt-payment-extension-ios.git (from 2.0.0).
- mParticle-Rokt (unchanged product name) now links mParticle-Rokt-Swift plus a small mParticle-Rokt-PaymentLinkage target that depends on RoktPaymentExtension, so the extension is available by default without renaming or changing the primary product.
- Adds mParticle-Rokt-No-Payments, a second library product that depends only on mParticle-Rokt-Swift (no RoktPaymentExtension / Stripe).
- Raises swift-tools-version to 5.9 for alignment with the payment extension toolchain.

- CocoaPods (mParticle-Rokt.podspec)

- Introduces Core (sources, privacy bundle, mParticle-Apple-SDK, RoktContracts, Rokt-Widget).
- Payments is the default_subspec: Core + RoktPaymentExtension ~> 2.0.
- No-Payments: Core only (pod 'mParticle-Rokt/No-Payments').

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ For each release, **Core** (main SDK) changes are listed first, followed by **Ki
 ##### Added
 
 - Pass through `handleURLCallback:` to `Rokt.handleURLCallback(with:)` on the Rokt SDK.
+- SwiftPM: declare `rokt-payment-extension-ios`; default product `mParticle-Rokt` links `RoktPaymentExtension`; optional product `mParticle-Rokt-No-Payments` omits it (no Stripe).
+- CocoaPods: default subspec `Payments` adds `RoktPaymentExtension`; subspec `No-Payments` is kit only (`Core`); subspec `Core` holds sources and base Rokt deps.
 
 ## [9.0.1] - 2026-04-22
 
--- a/Kits/rokt/rokt/Sources/mParticle-Rokt-PaymentLinkage/MPRoktPaymentExtensionLinkage.swift
+++ b/Kits/rokt/rokt/Sources/mParticle-Rokt-PaymentLinkage/MPRoktPaymentExtensionLinkage.swift
@@ -1,0 +1,5 @@
+import RoktPaymentExtension
+
+public enum MPRoktPaymentExtensionLinkage {
+    public static let isPaymentExtensionLinked = true
+}
--- a/Kits/rokt/rokt/Package.swift
+++ b/Kits/rokt/rokt/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.9
 
 import Foundation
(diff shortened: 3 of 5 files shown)


Kits/rokt/rokt/Sources/mParticle-Rokt-PaymentLinkage/MPRoktPaymentExtensionLinkage.swift
```diff
@@ -0,0 +1,5 @@
+import RoktPaymentExtension
+
+public enum MPRoktPaymentExtensionLinkage {
+    public static let isPaymentExtensionLinked = true
```
denischilik, line 4: Do we use isPaymentExtensionLinked anywhere?

denischilik (approved):

Kits/rokt/rokt/Sources/mParticle-Rokt-PaymentLinkage/MPRoktPaymentExtensionLinkage.swift
BrandonStalnaker: No, it really exists so the linkage target isn’t an “import-only” file with no symbols and cause I could see how it could be useful in the future.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ For each release, **Core** (main SDK) changes are listed first, followed by **Ki
 ##### Added
 
 - Pass through `handleURLCallback:` to `Rokt.handleURLCallback(with:)` on the Rokt SDK.
+- SwiftPM: declare `rokt-payment-extension-ios`; default product `mParticle-Rokt` links `RoktPaymentExtension`; optional product `mParticle-Rokt-No-Payments` omits it (no Stripe).
+- CocoaPods: default subspec `Payments` adds `RoktPaymentExtension`; subspec `No-Payments` is kit only (`Core`); subspec `Core` holds sources and base Rokt deps.
```
thomson-t, line 36: Can we Add some details in README as well?

jamesnrokt:
My main concern here is it could be pretty confusing given the two options

Screenshot 2026-05-01 at 4 40 17 PM

BrandonStalnaker:
Updated the README.md with clear instructions. Also beginning work on an alternate Rokt iOS SDK+ so we can choose between this implementation and it.

jamesnrokt:
Closing as solved by #766

PR #770, opened 2026-05-14
fix: throttle retriable transport upload failures

denischilik:
## Background
Some transport/network failures can produce excessive retry traffic when uploads cannot reach mParticle endpoints. This update introduces explicit transport-error detection and keeps throttling behavior aligned with existing retry flow.

## What Has Changed
- Added a Swift `MPTransportErrorDetector` to centralize retriable transport error classification.
- Applied transport-aware throttling in both message and alias upload retry paths.
- Kept existing 429/503 handling order while adding transport-failure throttling in the non-success retry branch.
- Added unit tests for transport error detection and wired them into the iOS test target.

## Screenshots/Video
N/A

## Checklist

### DIFF
--- a/mParticle-Apple-SDK-Swift/Sources/Network/MPNetworkCommunicationHelper.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Network/MPNetworkCommunicationHelper.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+@objc public final class MPNetworkCommunicationHelper: NSObject {
+    @objc(calculateRetryTimeForHeaders:)
+    public static func calculateRetryTime(for headers: NSDictionary) -> NSNumber {
+        let retryAfterDate = headers.retryDate()
+        let retryAfterSeconds = headers.retrySeconds()
+        let defaultRetryAfter: Double = 7200
+        let maxRetryAfter: Double = 86400
+
+        if let retryAfterDate {
+            let now = Date()
+            let retryAfter = min(retryAfterDate.timeIntervalSince1970 - now.timeIntervalSince1970, maxRetryAfter)
+            return NSNumber(value: retryAfter > 0 ? retryAfter : defaultRetryAfter)
+        }
+
+        if let retryAfterSeconds {
+            return NSNumber(value: min(retryAfterSeconds.doubleValue, maxRetryAfter))
+        }
+
+        return NSNumber(value: defaultRetryAfter)
+    }
+}
--- a/mParticle-Apple-SDK-Swift/Sources/Network/MPNetworkCommunicationHelper.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Network/MPNetworkCommunicationHelper.swift
@@ -3,8 +3,8 @@ import Foundation
 @objc public final class MPNetworkCommunicationHelper: NSObject {
     @objc(calculateRetryTimeForHeaders:)
     public static func calculateRetryTime(for headers: NSDictionary) -> NSNumber {
-        let retryAfterDate = headers.retryDate()
-        let retryAfterSeconds = headers.retrySeconds()
+        let retryAfterDate = headers.mp_retryDate()
+        let retryAfterSeconds = headers.mp_retrySeconds()
         let defaultRetryAfter: Double = 7200
         let maxRetryAfter: Double = 86400
 
--- a/mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
@@ -6,6 +6,8 @@ import Foundation
     private static let retryAfterSchedule: [Double] = [60, 300, 1800, 21600]
     private static var consecutiveTransportErrorCount = 0
     private static let backoffQueue = DispatchQueue(label: "com.mparticle.transport-error-backoff")
+    private static let semaphoreTimeoutErrorDomain = "com.mparticle"
+    private static let semaphoreTimeoutErrorCode = 0
 
     @objc(isRetriableTransportError:)
     public static func isRetriableTransportError(_ error: NSError?) -> Bool {
@@ -31,7 +33,8 @@ import Foundation
             }
         }
 
-        return false
+        return error.domain == semaphoreTimeoutErrorDomain
+            && error.code == semaphoreTimeoutErrorCode
     }
 
     @objc(calculateRetryTimeForTransportError)
--- a/mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
@@ -1,16 +1,16 @@
 import Foundation
 
 @objc public class MPTransportErrorDetector: NSObject {
-    private static let noConnectionErrorCode = 1
+    private static let maxRetryAfter: Double = 86400
+    private static let maxErrorCountBeforeMaxRetry = 5
+    private static let retryAfterSchedule: [Double] = [60, 300, 1800, 21600]
+    private static var consecutiveTransportErrorCount = 0
+    private static let backoffQueue = DispatchQueue(label: "com.mparticle.transport-error-backoff")
 
     @objc(isRetriableTransportError:)
     public static func isRetriableTransportError(_ error: NSError?) -> Bool {
         guard let error else {
             return false
-        }
-
-        if error.code == noConnectionErrorCode {
-            return true
         }
 
         if error.domain == NSURLErrorDomain {
@@ -31,6 +31,27 @@ import Foundation
             }
         }
 
-        return error.domain == "com.mparticle" && error.code == 0
+        return false
+    }
+
+    @objc(calculateRetryTimeForTransportError)
+    public static func calculateRetryTimeForTransportError() -> NSNumber {
+        return backoffQueue.sync {
+            consecutiveTransportErrorCount += 1
+
+            if consecutiveTransportErrorCount >= maxErrorCountBeforeMaxRetry {
+                return NSNumber(value: maxRetryAfter)
+            }
+
+            let scheduleIndex = max(0, consecutiveTransportErrorCount - 1)
+            return NSNumber(value: retryAfterSchedule[scheduleIndex])
+        }
+    }
+
+    @objc(resetTransportErrorCounter)
+    public static func resetTransportErrorCounter() {
+        backoffQueue.sync {
+            consecutiveTransportErrorCount = 0
+        }
     }
 }
--- a/mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
@@ -37,8 +37,7 @@ import Foundation
                 NSURLErrorSecureConnectionFailed,
                 NSURLErrorInternationalRoamingOff,
                 NSURLErrorDataNotAllowed,
-                NSURLErrorCallIsActive,
-                NSURLErrorAppTransportSecurityRequiresSecureConnection:
+                NSURLErrorCallIsActive:
                 return true
             default:
                 return false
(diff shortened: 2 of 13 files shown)


BrandonStalnaker (commented):
The code looks solid to me. I'm somewhat worried the change is a bit broad. I'm worried that for a few of these error types we would want to retry sooner than 2 hours from the failure (I think thats the minimum right). I'd like anothers opinion before I'll approve.
NSURLErrorCannotFindHost,
NSURLErrorCannotConnectToHost,
NSURLErrorNetworkConnectionLost,
NSURLErrorDNSLookupFailed,
NSURLErrorCannotLoadFromNetwork,
NSURLErrorSecureConnectionFailed,
NSURLErrorInternationalRoamingOff,
NSURLErrorDataNotAllowed,
NSURLErrorCallIsActive,
NSURLErrorAppTransportSecurityRequiresSecureConnection:

thomson-t (dismissed):

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
```diff
@@ -0,0 +1,36 @@
...
+                NSURLErrorDataNotAllowed,
+                NSURLErrorCallIsActive,
+                NSURLErrorAppTransportSecurityRequiresSecureConnection:
+                return true
+            default:
+                return false
+            }
+        }
+
+        return error.domain == "com.mparticle" && error.code == 0
```
jamesnrokt: We should extract these magic strings and number

Edit: Changed from Nit as a simple change in `MPConnector.m` could could an issue for us but the tests would still pass. We should have this and MPConnector.m share the same constant

mParticle-Apple-SDK/Network/MPNetworkCommunication.m
```diff
@@ -786,6 +795,10 @@ - (BOOL)performMessageUpload:(MPUpload *)upload {
 
     //5xx, 0, 999, -1, etc
     if (!isSuccessCode && !isInvalidCode) {
+        if ([self isRetriableTransportError:error]) {
+            MPILogWarning(@"Throttling uploads after transport error.");
+            [self throttleWithHTTPResponse:httpResponse uploadType:MPUploadTypeMessage];
```
jamesnrokt: Agree with Brandon. These retriable transport errors shouldn't be treated the same as a 429 or 503.

Those status codes are deliberate server-side backoffs to avoid DDoS'ing ourselves. Transport errors are transient client-side conditions that typically clear in seconds. With this approach, a user on the subway losing signal for 10 seconds would freeze uploads for 2 hours.

Can we use a separate, shorter window for transport errors e.g. 30s with exponential backoff, capped at a few minutes?

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
jamesnrokt: It looks like com.mparticle code 0 is the SDK's own semaphore timeout (`MPConnector.m:276/:321`), and the test naming confirms it's treated as a timeout. But NSURLErrorTimedOut isn't in the switch above and falls through to false.

May be mistaken but it looks like these are the same condition at different layers, the SDK timeout fires precisely when the URLSession callback didn't return in time. We should treat both the same way: either add NSURLErrorTimedOut to the retriable list, or drop this last return.

jamesnrokt (changes requested):
Requesting changes based on the above comments.

My core concern: as it stands, any retriable transport error puts the SDK into a 2-hour upload freeze with no recovery hook as minUploadDate is only checked, never cleared early.

Even a brief connectivity drop costs 2 hours of buffered events. Worse, because every client uses the same hardcoded default with no jitter, a brief endpoint blip triggers a synchronized thundering herd of requests hitting our servers 2 hours later.

mParticle-Apple-SDK/Network/MPNetworkCommunication.m
denischilik: Add different retry mechanism

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
```diff
@@ -0,0 +1,57 @@
...
+                NSURLErrorSecureConnectionFailed,
+                NSURLErrorInternationalRoamingOff,
+                NSURLErrorDataNotAllowed,
+                NSURLErrorCallIsActive,
+                NSURLErrorAppTransportSecurityRequiresSecureConnection:
+                return true
+            default:
+                return false
+            }
+        }
```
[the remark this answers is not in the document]
denischilik: fixed

mParticle-Apple-SDK/Network/MPNetworkCommunication.m
```diff
@@ -780,12 +770,19 @@ - (BOOL)performMessageUpload:(MPUpload *)upload {
...
+        [self throttleWithRetryAfter:retryAfter uploadType:MPUploadTypeMessage];
         return YES;
     }
 
     //5xx, 0, 999, -1, etc
     if (!isSuccessCode && !isInvalidCode) {
+        if ([self isRetriableTransportError:error]) {
+            MPILogWarning(@"Throttling uploads after transport error.");
+            NSTimeInterval retryAfter = [[MPTransportErrorDetector calculateRetryTimeForTransportError] doubleValue];
+            [self throttleWithRetryAfter:retryAfter uploadType:MPUploadTypeMessage];
```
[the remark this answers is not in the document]
denischilik: We should not reset the transport-error counter on non-2xx responses: receiving an HTTP response does not necessarily mean service health or delivery conditions have recovered, and resetting early would weaken backoff during ongoing degradation. The counter should reset only on successful uploads (2xx).

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+
+@objc public class MPTransportErrorDetector: NSObject {
+    private static let maxRetryAfter: Double = 86400
+    private static let maxErrorCountBeforeMaxRetry = 5
+    private static let retryAfterSchedule: [Double] = [60, 300, 1800, 21600]
```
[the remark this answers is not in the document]
denischilik: Updated to shorter intervals

mParticle-Apple-SDK/Network/MPNetworkCommunication.m
```diff
@@ -746,16 +732,20 @@ - (BOOL)performMessageUpload:(MPUpload *)upload {
...
+    NSError *error = response.error;
     NSHTTPURLResponse *httpResponse = response.httpResponse;
 
     NSInteger responseCode = [httpResponse statusCode];
     MPILogVerbose(@"Upload response code: %ld", (long)responseCode);
     BOOL isSuccessCode = responseCode >= 200 && responseCode < 300;
     BOOL isInvalidCode = responseCode != 429 && responseCode >= 400 && responseCode < 500;
+    if (isSuccessCode) {
+        [MPTransportErrorDetector resetTransportErrorCounter];
+    }
```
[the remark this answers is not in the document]
denischilik: We should not reset the transport-error counter on non-2xx responses: receiving an HTTP response does not necessarily mean service health or delivery conditions have recovered, and resetting early would weaken backoff during ongoing degradation. The counter should reset only on successful uploads (2xx).

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
denischilik: Extracted to constants

denischilik: return NSURLErrorTimedOut back to TransportError

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
```diff
@@ -0,0 +1,78 @@
...
+        if error.domain == NSURLErrorDomain {
+            switch error.code {
+            case NSURLErrorCannotFindHost,
+                NSURLErrorCannotConnectToHost,
+                NSURLErrorNotConnectedToInternet,
+                NSURLErrorNetworkConnectionLost,
+                NSURLErrorDNSLookupFailed,
+                NSURLErrorTimedOut,
+                NSURLErrorCannotLoadFromNetwork,
+                NSURLErrorSecureConnectionFailed,
```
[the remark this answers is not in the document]
denischilik: That is expected in this scenario

mParticle-Apple-SDK-Swift/Sources/Network/MPNetworkCommunicationHelper.swift
```diff
@@ -0,0 +1,23 @@
...
+
+        if let retryAfterDate {
+            let now = Date()
+            let retryAfter = min(retryAfterDate.timeIntervalSince1970 - now.timeIntervalSince1970, maxRetryAfter)
+            return NSNumber(value: retryAfter > 0 ? retryAfter : defaultRetryAfter)
+        }
+
+        if let retryAfterSeconds {
+            return NSNumber(value: min(retryAfterSeconds.doubleValue, maxRetryAfter))
+        }
```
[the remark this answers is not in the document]
denischilik: It is old logic i keep it the same

BrandonStalnaker (approved):

mParticle-Apple-SDK/Network/MPConnector.m
```diff
@@ -6,6 +6,12 @@
 #import "MPILogger.h"
 #import "mParticle.h"
 #import "MPURL.h"
+@import mParticle_Apple_SDK_Swift;
+
+@interface MPTransportErrorDetector (MPConnectorTimeout)
++ (NSString *)semaphoreTimeoutErrorDomain;
++ (NSNumber *)semaphoreTimeoutErrorCode;
+@end
```
jamesnrokt, line 14: Is this needed? `@import mParticle_Apple_SDK_Swift;` already imports the auto-generated Swift header which exposes `+ (NSString * _Nonnull)semaphoreTimeoutErrorDomain;` from  `@objc(semaphoreTimeoutErrorDomain) public static func`

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
+
+@objc public class MPTransportErrorDetector: NSObject {
+    private static let maxRetryAfter: Double = 300
+    private static let maxErrorCountBeforeMaxRetry = 5
+    private static let retryAfterSchedule: [Double] = [5, 15, 60, 120, 300]
+    private static var consecutiveTransportErrorCount = 0
```
jamesnrokt, line 7: Does there need to be some reset mechanic here for SDK init or similar? Wondering if it should also be split by host

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
```diff
@@ -0,0 +1,78 @@
...
+    private static let backoffQueue = DispatchQueue(label: "com.mparticle.transport-error-backoff")
+    private static let semaphoreTimeoutErrorDomainValue = "com.mparticle"
+    private static let semaphoreTimeoutErrorCodeValue = 0
```
jamesnrokt, line 10: Should we make these more strongly typed and specific to the use case? 0 feels easy to fall into as a default value and domain value could be more explicit generally.

jamesnrokt (approved):
Some comments but will leave up to you depending on urgency

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
denischilik: We initialize sdk just once, and because variable in memory it reset to 0 during app restart

mParticle-Apple-SDK-Swift/Sources/Network/MPTransportErrorDetector.swift
denischilik: I want to keep these constant the same like they were before, all logic can be addressed in refactor network layer lauchpad

PR #771, opened 2026-05-18
Brazekit not initialzied when loaded from cache during async call of MParticle.sharedInstance.Start()

ianhowertonmcd1:
Call kitInstance.start when loading a cached kit


BrandonStalnaker:
Thank you for bringing this issue to our attention. I believe it was obscured by many of our other clients initializing Braze manually and most other kits no longer using the 'start' paradigm. We've implemented a change in the same vein as your fix and merged it to go out in our next release.

PR #805, opened 2026-08-06
fix: Update Binary Framework Generation to Dynamic for Appsflyer Kit

BrandonStalnaker:
## Background
- A customer integrating the mParticle Apple SDK and the AppsFlyer kit via prebuilt xcframeworks (.binaryTarget) crashed at launch after upgrading, with duplicate ObjC class warnings for core SDK types (MParticle, MPIdentityApi, etc.).
- Root cause: the published mParticle_AppsFlyer.xcframework was archived from the SPM scheme, which links mParticle-Apple-SDK and AppsFlyerLib-Static as static libraries. Those were absorbed into the kit binary, so an app that also linked the core SDK xcframework ended up with two copies of every class and split singleton state.
- This has been true of every AppsFlyer kit xcframework since v9.0.0; binary consumers appear to be newly hitting it. The existing header smoke test only ran clang -fsyntax-only and could not catch duplicate symbols. SPM and CocoaPods source integrations were unaffected.

## What Has Changed
- Added a mParticle-AppsFlyer-Dynamic Xcode target/scheme that compiles the same kit sources but links -framework mParticle_Apple_SDK and -framework AppsFlyerLib dynamically, so the distributable binary defines only MPKitAppsFlyer.
- Introduced reusable Scripts/build_kit_xcframework.sh, driven by a dynamic_xcframework block in Kits/matrix.json (scheme, module, partner dependency URL/package_identity). AppsFlyer is the first kit wired through it; other kits can reuse the same script later.
- Release CI now routes kits with dynamic_xcframework through that script instead of the generic SPM archive path.
- Added Scripts/verify_kit_xcframework_symbols.sh to fail the build when a kit xcframework redefines ObjC/Swift symbols from the core SDK or partner frameworks, and updated the import smoke job to build the real distributable and run that check.
- Documented binary-consumer requirements in the AppsFlyer kit README: embed mParticle_AppsFlyer.xcframework, mParticle_Apple_SDK.xcframework, and the dynamic AppsFlyerLib.xcframework together.

## Checklist

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -151,6 +151,14 @@ jobs:
 
           mkdir -p archives xcframeworks
 
+          # Kits with dynamic_xcframework link partner SDKs at runtime instead of
+          # absorbing static SPM products into the kit binary.
+          if echo '${{ toJson(matrix.kit.dynamic_xcframework) }}' | jq -e . >/dev/null 2>&1; then
+            chmod +x ./Scripts/build_kit_xcframework.sh
+            OUTPUT_DIR="$PWD/xcframeworks" ./Scripts/build_kit_xcframework.sh "${{ matrix.kit.name }}"
+            exit 0
+          fi
+
           SKIP_XCF="${{ matrix.kit.skip_xcframework }}"
           if [ "$SKIP_XCF" = "true" ]; then
             MODULE=$(echo "$SCHEMES_JSON" | jq -r '.[0].module')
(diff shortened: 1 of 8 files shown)


denischilik (approved):
LGTM — fixes the real root cause for AppsFlyer binary consumers.

Could you either roll the same `dynamic_xcframework` packaging out to the other kits, or add a task to the spreadsheet so we track that follow-up?

thomson-t (approved):

PR #814, opened 2026-08-14
fix: notify callers when a Rokt placement cannot be requested

jamesnrokt:
## Background

- `MPRokt.selectPlacements:attributes:embeddedViews:config:onEvent:` resolves the dashboard attribute mapping via `getRoktPlacementAttributesMapping` before forwarding to the Rokt kit. That method returns `nil` whenever the Rokt kit configuration is not present in `MPKitContainer.originalConfig` — most commonly because remote configuration has not been applied yet, and also when the kit is not enabled for the workspace.
- In that case the SDK logged a warning and returned. The caller's `onEvent` handler was never invoked, so partners waiting on a placement lifecycle event (loading indicator, spinner, checkout continuation) were left waiting on a callback that would never arrive, with nothing to distinguish "not configured" from "still loading".
- `MPRokt.selectShoppableAds:attributes:config:onEvent:` had the same behaviour on the same code path.

## What Has Changed

- Both `selectPlacements` and `selectShoppableAds` now deliver `RoktEvent.PlacementFailure` (`RoktPlacementFailure`, carrying the requested placement identifier) to the caller's `onEvent` handler when the Rokt kit configuration is unavailable. The existing warning log is unchanged.
- The event is delivered on the main queue, since callers drive UI from this handler. Nothing is delivered when `onEvent` is `nil` (including the convenience `selectPlacements:attributes:` overload).
- `RoktPlacementFailure` is the event the Rokt SDK already emits when a placement cannot be displayed, so partner event handlers need no new cases.
- No API surface change: this is behaviour on an existing callback.
- Added unit tests covering the placement-failure callback for both entry points, a nil-`onEvent` regression test, and a CHANGELOG entry.

## Scope

- This PR only guarantees that the caller is always notified. It intentionally does not change _when_ the call is dropped. A related gap remains open and is not addressed here: `MPKitContainer.forwardSDKCall` discards a forwarded call when the Rokt kit is registered and configured but has not finished starting, since the kit only sets `started` from the asynchronous `Rokt.init` callback.

## Checklist

## Additional Notes

- Partners who currently treat "no callback" as an implicit failure signal will now receive an explicit `RoktPlacementFailure`. This is additive, but it is the first time this code path emits an event at all, so it is worth calling out in release notes.

### DIFF
--- a/mParticle-Apple-SDK/MPRokt.m
+++ b/mParticle-Apple-SDK/MPRokt.m
@@ -115,6 +115,7 @@ static const NSInteger kMPRoktKitId = 181;
             });
         } else {
             MPILogWarning(@"MPRokt selectPlacements not performed - Rokt Kit not configured. Check with your Rokt representative to ensure the kit is enabled.");
+            [self notifyPlacementFailure:identifier onEvent:onEvent];
         }
     }];
 }
@@ -383,6 +384,7 @@ static const NSInteger kMPRoktKitId = 181;
             });
         } else {
             MPILogWarning(@"MPRokt selectShoppableAds not performed - Rokt Kit not configured. Check with your Rokt representative to ensure the kit is enabled.");
+            [self notifyPlacementFailure:identifier onEvent:onEvent];
         }
     }];
 }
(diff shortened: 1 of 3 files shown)


mParticle-Apple-SDK/MPRokt.m
```diff
@@ -114,6 +114,7 @@ - (void)selectPlacements:(NSString *)identifier
             });
         } else {
             MPILogWarning(@"MPRokt selectPlacements not performed - Rokt Kit not configured. Check with your Rokt representative to ensure the kit is enabled.");
+            [self notifyPlacementFailure:identifier onEvent:onEvent];
```
BrandonStalnaker, line 118: Good catch. I wish we could do more of this logic in the actual kit than here in the mparticle side. This attributte mapping block means the kit forwarding retry logic effectively doesn't work for rokt selectPlacements.

BrandonStalnaker (dismissed):
solid improvement

thomson-t (approved):

BrandonStalnaker (approved):

mParticle-Apple-SDK/MPRokt.m
jamesnrokt: Let's add that to the backlog

PR #830, opened 2026-08-24
refactor: move NSDictionary+MPCaseInsensitive logic to Swift

nickolas-dimitrakas:
## Background

Stacked on #829, which added the characterisation tests this change is validated against. Those tests were written and landed against the existing Objective-C implementation first, so this PR is a pure implementation swap: the tests do not change here, only what they run against.

## Public API

`mParticle-Apple-SDK/Include/NSDictionary+MPCaseInsensitive.h` is **byte-identical** to the base commit; it does not appear in this diff at all. The category, its three selectors and their nullability are unchanged, so apps and the shipped Braze 12/13/14 and Optimizely 4/5 kit binaries that call `transformValuesToString` see no difference.

## What changed

Two files. No project file changes, because `mParticle-Apple-SDK-Swift` is a synchronized root group and the `.m` already exists.

- New `mParticle-Apple-SDK-Swift/Sources/Utils/NSDictionary+MPCaseInsensitive.swift`.
- `NSDictionary+MPCaseInsensitive.m` keeps the category, the `@try`/`@catch`, the `MPILogError` calls and the `NSAssert`, and forwards the actual work to Swift.

The Swift side uses `mp`-prefixed selectors so it cannot collide with or shadow the public ones.

## Three things worth reviewing closely

**Boolean encoding.** The original relied on pointer identity against `kCFBooleanTrue`/`kCFBooleanFalse` to emit `"true"`/`"false"` while leaving `@1` as `"1"`. Swift cannot do that comparison meaningfully once values cross the bridge, so this uses `CFGetTypeID(number) == CFBooleanGetTypeID()`, which distinguishes `__NSCFBoolean` from `__NSCFNumber`. This is the single most breakable behaviour in the file and it is covered directly.

**Why the transformer has a separate support check.** Both `MPDateFormatter.string(fromDateRFC3339:)` and UTF-8 decoding return optionals, so a *supported* value can still convert to nil. Today that drops the key silently, whereas an *unsupported* type logs and trips `NSAssert`. A single nil-returning function cannot express both, so the transformer exposes `isSupportedAttributeValue:` alongside `transformedValueForAttribute:` and the facade decides. The two functions mirror each other branch for branch.

**Why the assert stayed in Objective-C.** Swift's `assert` traps rather than raising a catchable `NSException`, which would have turned a recoverable condition into a hard crash and broken the existing behaviour. Keeping the log and `NSAssert` in the `.m` preserves it exactly.

## Testing

- All 19 characterisation tests from #829 pass unchanged against the Swift implementation.
- Targeted run of every suite that exercises `transformValuesToString`, `MPMessageBuilderTests`, `MPCommerceEventTests`, `MPKitContainerTests`, `MPBase_Attribute_Event_ProjectionTests`, plus `MPRoktTests` and `MPStateMachineTests`: 210 passed, 1 failed.
- Full suite: 955 passed, 17 failed.

None of the failures relate to this change. Across four runs on this stack the failing set rotated between a single `MPRoktTests` crash, 17 `MPRoktTests` async waits exceeding a 0.2 second timeout plus an `MPStateMachineTests` thread-safety crash, and a single `MPKitContainerTests/testActiveKitsRegistryThreadSafety`. They are all thread-safety or short-timeout tests and they move between runs under machine load. The untouched base commit `86fa6475` reproduces the same class of failure on its own, so this is pre-existing. It may be worth tightening those timeouts independently.

- `trunk check` clean via the pre-commit hook.


nickolas-dimitrakas:
Kit verification done locally, since both affected kit families call `transformValuesToString` directly.

Built each against this branch with `USE_LOCAL_VERSION=1`, which repoints the kit package at the local SDK:

- `Kits/optimizely/optimizely-5` — BUILD SUCCEEDED
- `Kits/braze/braze-14` — BUILD SUCCEEDED

Both logs confirm resolution to the local worktree rather than a released version (`mParticle-Apple-SDK: ... @ local`).

mParticle-Apple-SDK-Swift/Sources/Utils/NSDictionary+MPCaseInsensitive.swift
```diff
@@ -0,0 +1,81 @@
...
+            return true
+        }
+        if value is NSArray {
+            return true
+        }
+        if value is NSNull {
+            return true
+        }
+        return false
+    }
```
thomson-t, line 33: How about something like this
``` swift
@objc(isSupportedAttributeValue:)
public static func isSupportedValue(_ value: Any) -> Bool {
    switch value {
    case is String, is NSNumber, is Date, is NSDictionary, is NSArray, is NSNull:
        return true
    case let data as Data:
        return !data.isEmpty
    default:
        return false
    }
}
```

nickolas-dimitrakas: Taken, thanks. Checked it was behaviour preserving before applying: `Data` does not match any of `String`, `NSNumber`, `Date`, `NSDictionary`, `NSArray` or `NSNull`, so it always reaches the `case let data as Data` arm regardless of arm order. Reads much better.

Kept the `Data` arm separate since its support is conditional on being non-empty, and reworded the doc comment, which previously claimed the function mirrored the branch *order* of `transformedValue(for:)`. That is no longer literally true; the invariant that actually matters is that the two agree on the supported type *set*.

19 of 19 characterisation tests still pass.

thomson-t (dismissed):

BrandonStalnaker (dismissed):

thomson-t (approved):

PR #838, opened 2026-08-24
fix: keep MPRoktTests async work inside its mocks' lifetime

nickolas-dimitrakas:
## Background

The `native-tests / native-unit-tests (tvOS, mParticle-Apple-SDK)` job fails on essentially every run and has been blocking the stacked migration PRs. It reads as flaky — a different test dies each time — but it is one deterministic defect.

`MPRoktTests` dispatches forwards onto `[MParticle messageQueue]` fire-and-forget. A test can return with its block still queued. `tearDown` then calls `stopMocking` on the `MParticle`/`MPRokt` partial mocks, the queued block lands on a torn-down mock, and OCMock raises **on the message queue**, where XCTest has no handler for it. The uncaught exception aborts the test process, so whichever test is running at that moment never reports a result and the job exits 65.

The suite already documented this, at `MPRoktTests.m:97` — but the mitigation, `drainMessageQueue`, was called from **1** of 61 tests, deliberately kept out of `tearDown` on cost grounds.

## Why tvOS and not iOS

Same commits pass on iOS and fail on tvOS. The tvOS simulator is slow enough that the 0.2s async waits expire before the work lands, so tests return with blocks still queued. On iOS the work finishes inside the budget and nothing leaks.

## Evidence it is pre-existing

Same crash class, five different messages, on branches with no common feature work:

| Run | Uncaught exception |
|---|---|
| `main` @ 2026-08-21 | `No partial mock for object 0x…` |
| #834 (merged) | `+[MParticle sharedInstance]: unrecognized selector sent to class` |
| #830 | `Cannot use mock object OCPartialMockObject(MParticle)…` |
| #833 attempt 5 | `-[MPRokt confirmSandboxAttribute:]: unrecognized selector` |
| #833 attempt 6 | `-[MParticle ocmock_replaced_logLevel]: unrecognized selector` |

The orphaned test varies with timing — `testApiKeySecretThreadSafety`, `testActiveKitsRegistryThreadSafety`, `testSwitchWorkspaceSideloadedKits` — and is a bystander, not the culprit. #833 attempt 4 is the clearest case: zero timeouts, 888 of 889 passed, still exit 65.

## What changed

- `drainMessageQueue` moved into `tearDown`, before any `stopMocking`, so it covers every test rather than the one known to dispatch. The helper is unchanged: it enqueues a sentinel on the serial queue and waits, bounded at 5s.
- The 27 `waitForExpectationsWithTimeout:` and 5 `verifyWithDelay:` sites unified on `kAsyncWaitTimeout` (5.0s). Both XCTest and OCMock return as soon as the condition holds, so a larger budget costs nothing on the passing path — it only stops the slower simulator failing on wall clock.
- Removed the now-redundant explicit `drainMessageQueue` call and the stale comment on the one wait that used to differ.

Test-only change. No SDK source touched, no public API affected.

## Testing

- iOS `mParticle-Apple-SDK`, local: see comment below.
- tvOS is the case that matters and I could not run it locally — this machine has the tvOS 26.2 SDK but not its simulator runtime. **CI on this PR is the real verification.** Worth watching the tvOS job across a couple of re-runs before trusting it.

## Note for reviewers

This does not unblock the open stack directly — those branches are based on older commits and will need a rebase to pick it up.

### DIFF
--- a/UnitTests/ObjCTests/MPRoktTests.m
+++ b/UnitTests/ObjCTests/MPRoktTests.m
@@ -14,6 +14,11 @@
 // Rokt kit identifier for testing
 static NSNumber * const kTestRoktKitId = @181;
 
+/// Budget for the async waits below. Both XCTest and OCMock return as soon as the condition
+/// holds, so a generous value costs nothing when a test passes; it only stops the slower tvOS
+/// simulator from failing on wall clock while the work is still in flight.
+static const NSTimeInterval kAsyncWaitTimeout = 5.0;
+
 // Test helper class that simulates a kit with getSessionId and handleURLCallback methods
 @interface MPRoktTestKitInstance : NSObject
 @property (nonatomic, copy) NSString *sessionIdToReturn;
@@ -80,6 +85,7 @@ static NSNumber * const kTestRoktKitId = @181;
 }
 
 - (void)tearDown {
+    [self drainMessageQueue];
     self.rokt = nil;
     [self.mockRokt stopMocking];
     [self.mockInstance stopMocking];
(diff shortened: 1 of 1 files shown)


thomson-t (approved):

nickolas-dimitrakas:
Closing this as #837 is the better fix

PR #844, opened 2026-08-25
refactor: migrate identity request/filter/user logic to Swift

thomson-t:
## What

Migrates the **pure logic** of the public identity types behind their unchanged ObjC wrappers into Swift — the tracking-sheet item _"Identity request/filter/user (not the full API) — class extensions of public types"_ (MOD-07, scoped to exclude the full API). Follows the established two-stage wrapper model: logic → Swift helper, ObjC `.m` delegates, `.h`/selectors untouched. These are public types, so the ObjC wrappers stay **permanently**.

### Extracted logic

| ObjC type | Swift helper | What moved |
|---|---|---|
| `MPIdentityApiRequest` | `MPIdentityApiRequestStoragePRIVATE` | identity storage, `setIdentity` null/empty rules, typed accessors (`email`/`customerId`/`emailSha256`/`mobileSha256`) |
| `FilteredMParticleUser`, `FilteredMPIdentityApiRequest` | `MPIdentityFilteringPRIVATE` | the **duplicated** user-identity + user-attribute kit filters, consolidated into one shared helper |
| `MParticleUser.identities` | `MPIdentityFilteringPRIVATE` | stored-array → dictionary parse |

SDK-type access (data-plan filter, state machine, `MPIHasher`, kit config, the ATT strip) stays in ObjC and is passed across the boundary as Foundation values plus an `isBlocked` block — the Swift module never sees an SDK type (it cannot import the ObjC module).

### Out of scope ("not the full API")

`MPIdentityApi` (HTTP request orchestration) and `MPIdentityDTO` (HTTP body / field mapping) — those are the "full API" and a later slice.

## Validation

- ✅ **ABI guard** (`Tools/abi-guard.sh check`) — zero public-ABI diff
- ✅ **trunk check** — no issues
- ✅ **iOS build + tests** (iPhone 16 Pro, iOS 18.1):
  - Swift mirror tests: `MPIdentityApiRequestStorageTests` + `MPIdentityFilteringTests` — 18/18
  - ObjC contract tests (kept, unchanged): `MPIdentityApiRequestTests` 12/12, `MPKitAPITests` 4/4 (kit filter contract), `MParticleUserTests` 6/6, `MPIdentityTests` 32/32
 
## Notes for reviewers

- New `Test/` files are registered in **both** `PBXFileSystemSynchronizedBuildFileExceptionSet`s (removed from the framework target, added to the test target) — the only `project.pbxproj` change. `Sources/` files need no pbxproj edit.
- Minor behavior improvement: `MParticleUser.identities` now skips malformed persisted entries (missing `i`/`n`) instead of crashing on `setObject:nil` — safer on bad input, no change for well-formed data.


mParticle-Apple-SDK-Swift/Sources/Identity/MPIdentityFiltering.swift
```diff
@@ -0,0 +1,87 @@
...
+    /// Identity types at or above this raw value are device identities (IDFA,
+    /// IDFV, push token, application stamp) that are never forwarded to kits.
+    private static let firstDeviceIdentityType = MPIdentitySwift.iosAdvertiserId.rawValue
+
+    /// Parse the persisted user-identity array (`[{"n": type, "i": value}]`)
+    /// into a `type → value` map. Mirrors the array→dictionary transform in
+    /// `MParticleUser.identities`; the device-privacy (ATT) strip stays in ObjC
+    /// because it reads the state machine.
+    @objc(userIdentitiesFromStoredArray:)
+    public func userIdentities(fromStoredArray array: [[AnyHashable: Any]]?) -> [NSNumber: String] {
```
BrandonStalnaker, line 23: Returned filter maps are no longer NSMutableDictionary. The old code returned a live NSMutableDictionary typed as NSDictionary *. Swift [NSNumber: Any] / [String: Any] bridges as immutable. Kits I checked copy or only read (Braze, Kochava, Firebase), so this is probably fine. The only real issue is that it's an observable change. Returning NSMutableDictionary from the helper would keep the old object identity.

BrandonStalnaker (dismissed):
Just one maybe issue

mParticle-Apple-SDK-Swift/Sources/Identity/MPIdentityFiltering.swift
nickolas-dimitrakas: +1

thomson-t: Good catch — fixed in 32ea9ae. The filtered accessors now `mutableCopy` the Swift result at the ObjC boundary, so `FilteredMParticleUser.userIdentities`/`.userAttributes` and `FilteredMPIdentityApiRequest.userIdentities` return an `NSMutableDictionary` again (matching the pre-migration return type). Kept the mutability handling in the ObjC wrapper rather than the Swift helper so the helper stays Foundation-value-only. Added mutability assertions on both the empty and filtered paths (`MParticleUserTests`, `MPKitAPITests`).

BrandonStalnaker (approved):

PR #859, opened 2026-08-26
feat: Migrate Identity to Swift Implementation

BrandonStalnaker:
## What

Moves identity HTTP DTO mapping, alias/user-sort helpers, and user-identity storage into the Foundation-only Swift module while keeping **`MPIdentityDTO`, `MPIdentityApi`, and `MParticleUser` as Objective-C facades**. Public headers and selectors are unchanged so apps and kits are not broken.

Unlike [internal leaf types that can be deleted](https://github.com/mParticle/mparticle-apple-sdk/pull/857), these classes are public / kit-facing, so they stay in the ObjC target with a Swift `*PRIVATE` implementation behind them — the same pattern as identity request/filter (`#844`) and `MPStateMachine` (`#854`).

## How the ObjC-module dependency cycle is handled

Swift cannot import the ObjC module. Logic that does not need `MParticle`, persistence, kits, or UIKit lives in Swift; ObjC `.m` files marshal and forward:

- `MPIdentityHTTPIdentitiesPRIVATE` / `MPIdentityHTTPRequestBuilderPRIVATE` / `MPIdentityHTTPIdentityChangePRIVATE` — identity type ↔ wire-string mapping, ATT IDFA gating, request/success dictionaries, `NSNull` for missing change values (and for identity values that were `NSNull` in the input, matching cache hashing).
- `MPIdentityAliasPlanPRIVATE` / `MPIdentityApiLogicPRIVATE` — alias MPID/date-window validation, last-seen sort indexes, modify-change parsing, HTTP error fields.
- `MPIdentityUserStoragePRIVATE` — first/last-seen dates, stored identity-array add/replace/remove, empty-attribute skip, audience-disabled error.

**Left in ObjC (still coupled to the rest of the SDK):** identify-request construction (persisted MPID, `MPDevice` vendor ID, DAS, iOS push token), environment string from `MParticle.sharedInstance`, identify/login/logout/modify networking, kit forwarding, MPID change, persistence, `currentUser` locking, and session `sessionUserIds` merge (`componentsSeparatedByString` on `""` is `[""]`; that behavior stays on the ObjC side).

## Behavior / API

- Public class names, headers, and selectors are unchanged.
- Identity HTTP dictionaries keep the same keys, nil vs `NSNull` semantics, iOS-only `push_token`, and ATT skip of IDFA when status is present and not authorized.
- Alias requests still reject missing/zero/equal MPIDs, clamp start when `usedFirstLastSeen` is set, and fill missing dates from the alias max window.

## Validation

- `trunk check`: clean for the identity Swift/ObjC sources (pre-existing markdown lint only in `.cursor/plans`).
- Build `mParticle-Apple-SDK` (iOS + tvOS): **succeeded**.
- Static analyzer (`mParticle-Apple-SDK`, Debug, iPhone simulator): **ANALYZE SUCCEEDED**, no diagnostics.
- **ObjC identity contract:** `MPIdentityTests`, `MParticleUserTests`, `MPIdentityCachingTests` (including `NSNull` google identity in identify-request hashing).
- New `MPIdentityDTOTests.swift`: type round-trip, ATT IDFA skip, `NSNull` identities/changes, alias plan, last-seen indexes, modify-change parse, identity-array mutation, empty-attribute skip.

### DIFF
--- a/UnitTests/ObjCTests/MPIdentityTests.m
+++ b/UnitTests/ObjCTests/MPIdentityTests.m
@@ -1107,4 +1107,22 @@ typedef NS_ENUM(NSUInteger, MPIdentityRequestType) {
     XCTAssertEqualObjects([MParticle sharedInstance].identity.deviceApplicationStamp, expected);
 }
 
+- (void)testIdentityChangePlainInitProducesNullValues {
+    MPIdentityHTTPIdentityChange *change = [[MPIdentityHTTPIdentityChange alloc] init];
+    NSMutableDictionary *dictionary = [change dictionaryRepresentation];
+    XCTAssertEqualObjects(dictionary[@"old_value"], [NSNull null]);
+    XCTAssertEqualObjects(dictionary[@"new_value"], [NSNull null]);
+    XCTAssertNil(dictionary[@"identity_type"]);
+}
+
+- (void)testHTTPIdentitiesPreservesNonStringValues {
+    MPIdentityHTTPIdentities *identities = [[MPIdentityHTTPIdentities alloc] initWithIdentities:@{
+        @(MPIdentityGoogle): @12345,
+        @(MPIdentityOther): [NSNull null]
+    }];
+    NSDictionary *dictionary = [identities dictionaryRepresentation];
+    XCTAssertEqualObjects(dictionary[@"google"], @12345);
+    XCTAssertEqualObjects(dictionary[@"other"], [NSNull null]);
+}
+
 @end
--- a/mParticle-Apple-SDK-Swift/Sources/Identity/MPIdentityApiLogic.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Identity/MPIdentityApiLogic.swift
@@ -1,0 +1,109 @@
+import Foundation
+
+@objc public final class MPIdentityAliasPlanPRIVATE: NSObject {
+    @objc public let isValid: Bool
+    @objc public let startTime: Date?
+    @objc public let endTime: Date?
+    @objc public let shouldWarnDateOrder: Bool
+
+    @objc public init(isValid: Bool, startTime: Date?, endTime: Date?, shouldWarnDateOrder: Bool) {
+        self.isValid = isValid
+        self.startTime = startTime
+        self.endTime = endTime
+        self.shouldWarnDateOrder = shouldWarnDateOrder
+        super.init()
+    }
+
+    @objc(planWithSourceMPID:destinationMPID:startTime:endTime:usedFirstLastSeen:aliasMaxWindow:)
+    public static func plan(
+        sourceMPID: NSNumber?,
+        destinationMPID: NSNumber?,
+        startTime: Date?,
+        endTime: Date?,
+        usedFirstLastSeen: Bool,
+        aliasMaxWindow: NSNumber?
+    ) -> MPIdentityAliasPlanPRIVATE {
+        let source = sourceMPID?.int64Value ?? 0
+        let destination = destinationMPID?.int64Value ?? 0
+        if sourceMPID == nil || destinationMPID == nil || source == 0 || destination == 0 || sourceMPID == destinationMPID {
+            return MPIdentityAliasPlanPRIVATE(isValid: false, startTime: startTime, endTime: endTime, shouldWarnDateOrder: false)
+        }
+
+        let maxDaysAgo = aliasMaxWindow?.doubleValue ?? 90
+        let oldestAllowableDate = Date(timeIntervalSinceNow: -60 * 60 * 24 * maxDaysAgo)
+        var adjustedStart = startTime
+        var adjustedEnd = endTime
+
+        if usedFirstLastSeen, let startTime, startTime.compare(oldestAllowableDate) == .orderedAscending {
+            adjustedStart = oldestAllowableDate
+        }
+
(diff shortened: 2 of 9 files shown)


nickolas-dimitrakas (changes requested):
Audited the ObjC→Swift logic diff. Constants, `@objc` selector names, `MPIdentitySwift`/`MPATTAuthorizationStatusSwift` raw values, the alias clamp-then-fill ordering, `NSNull` passthrough into `known_identities`, and the pbxproj `membershipExceptions` all check out. Two things to fix first, both in the DTO facades.

### 1. `MPIdentityHTTPIdentityChange` is missing an `-init` override

`MPIdentityHTTPIdentities` got an `init` that allocates its `implementation`; `MPIdentityHTTPIdentityChange` didn't. So `[[MPIdentityHTTPIdentityChange alloc] init]` now leaves `implementation` nil — every getter returns nil and `dictionaryRepresentation` returns **nil**, which crashes the modify path:

```objc
NSDictionary *changeDictionary = [obj dictionaryRepresentation];
[identityChanges addObject:changeDictionary];   // addObject:nil → NSInvalidArgumentException
```

Pre-migration, a plain-init'd change produced a valid `{old_value: null, new_value: null}`. Nothing in the SDK plain-inits it today (all three sites use `initWithOldValue:`), but it's a public header class with readwrite properties, so a kit or test can. Same one-liner as the sibling class:

```objc
- (instancetype)init {
    self = [super init];
    if (self) {
        _implementation = [[MPIdentityHTTPIdentityChangePRIVATE alloc] initWithOldValue:nil value:nil identityType:nil];
    }
    return self;
}
```

### 2. Identity values that aren't `String` are now silently dropped

The old ObjC assigned `obj` straight through regardless of class. `apply(_:to:key:)` in `MPIdentityDTO.swift` only writes on `value as? String` — anything else leaves the property untouched *and* doesn't register in `nullIdentityKeys`, so the key vanishes from `known_identities` entirely.

Worth more than a type nit because the public header declares `NSDictionary<NSNumber*, NSObject*> *identities`, so non-strings are inside the declared contract. And `known_identities` feeds the identity cache key (`MPIdentityCaching.m:207`), so a dropped value both never reaches the server and shifts the sha256 — silent cache divergence rather than a visible failure. `MPIdentityApiRequestStorage` only stores `String`/`NSNull` today, so it isn't reachable through the public setters, but this used to be a passthrough. Keep the passthrough, or at minimum log the drop.

BrandonStalnaker:
> Audited the ObjC→Swift logic diff. Constants, `@objc` selector names, `MPIdentitySwift`/`MPATTAuthorizationStatusSwift` raw values, the alias clamp-then-fill ordering, `NSNull` passthrough into `known_identities`, and the pbxproj `membershipExceptions` all check out. Two things to fix first, both in the DTO facades.
> 
> ### 1. `MPIdentityHTTPIdentityChange` is missing an `-init` override
> `MPIdentityHTTPIdentities` got an `init` that allocates its `implementation`; `MPIdentityHTTPIdentityChange` didn't. So `[[MPIdentityHTTPIdentityChange alloc] init]` now leaves `implementation` nil — every getter returns nil and `dictionaryRepresentation` returns **nil**, which crashes the modify path:
> 
> ```objc
> NSDictionary *changeDictionary = [obj dictionaryRepresentation];
> [identityChanges addObject:changeDictionary];   // addObject:nil → NSInvalidArgumentException
> ```
> 
> Pre-migration, a plain-init'd change produced a valid `{old_value: null, new_value: null}`. Nothing in the SDK plain-inits it today (all three sites use `initWithOldValue:`), but it's a public header class with readwrite properties, so a kit or test can. Same one-liner as the sibling class:
> 
> ```objc
> - (instancetype)init {
>     self = [super init];
>     if (self) {
>         _implementation = [[MPIdentityHTTPIdentityChangePRIVATE alloc] initWithOldValue:nil value:nil identityType:nil];
>     }
>     return self;
> }
> ```
> 
> ### 2. Identity values that aren't `String` are now silently dropped
> The old ObjC assigned `obj` straight through regardless of class. `apply(_:to:key:)` in `MPIdentityDTO.swift` only writes on `value as? String` — anything else leaves the property untouched _and_ doesn't register in `nullIdentityKeys`, so the key vanishes from `known_identities` entirely.
> 
> Worth more than a type nit because the public header declares `NSDictionary<NSNumber*, NSObject*> *identities`, so non-strings are inside the declared contract. And `known_identities` feeds the identity cache key (`MPIdentityCaching.m:207`), so a dropped value both never reaches the server and shifts the sha256 — silent cache divergence rather than a visible failure. `MPIdentityApiRequestStorage` only stores `String`/`NSNull` today, so it isn't reachable through the public setters, but this used to be a passthrough. Keep the passthrough, or at minimum log the drop.

1. Missing -init on MPIdentityHTTPIdentityChange
Plain [[MPIdentityHTTPIdentityChange alloc] init] now allocates the Swift implementation, so dictionaryRepresentation is {old_value: null, new_value: null} instead of nil (which would have crashed addObject: on modify).

2. Non-string identity values were dropped
apply now stores any NSObject the same way the old ObjC ivars did (NSString, NSNull, NSNumber, nested dictionaries, etc.). Those values stay in known_identities and in the identity cache key.

Added ObjC and Swift tests for both.

nickolas-dimitrakas (approved):
LGTM

thomson-t (approved):

PR #862, opened 2026-08-26
refactor: move scene delegate URL mapping to Swift

nickolas-dimitrakas:
Stacked on #861, which is stacked on #860. Last of three — this completes the
app-lifecycle + kit-forwarding area.

## What moves

`SceneDelegateHandler` is mostly logging, and logging is exactly what was never
under test. `MPSceneDelegateLogic` now owns:

- the `UIApplicationOpenURLOptionsSourceApplicationKey` options mapping,
- the `NSUserActivityTypeBrowsingWeb` check,
- and the assembly of both sets of log lines.

**Swift composes the strings; the `.m` still emits them through `MPILogDebug`.**
That is deliberate: `MPILogDebug` is a macro that captures the call site and reads
`MParticle`'s log level, so it cannot move. Composing in Swift and emitting in
ObjC makes the string building testable while leaving console output
byte-identical — no change to log level resolution, no change to formatting.

## The public wrapper stays, permanently

`Include/SceneDelegateHandler.h` is a public header, exported from
`Include/mParticle.h`, and the baseline records both `@interface
SceneDelegateHandler` and `@protocol OpenURLHandlerProtocol`. Per
`CONVERSION-RECIPE.md` a type with an `Include/` header keeps its ObjC wrapper —
deleting it would turn the class into a Swift `@objc` class, a module-identity
change that can break partner kits even with an identical name. So the class, its
initialiser and its two methods are untouched; only the body of the `.m` changed.

`Include/` has a zero diff and `abi-guard.sh check` is clean.

## Three behaviours preserved

1. **A missing source application yields an empty dictionary, not nil.** The ObjC
   version always handed `openURL:options:` an `NSMutableDictionary`, empty if
   there was nothing to put in it.
2. **A nil title or webpage URL logs as empty, not `(null)`.** The original used
   `?: @""` on both, which differs from the `%@` formatting of the other fields.
3. **The event-attribution line is absent, not empty, below iOS 14.5.** It sits
   behind an `@available` check, so the line simply is not emitted there — the
   Swift signature takes it pre-resolved as an optional to keep that shape.

## Gate

| Item | Result |
|---|---|
| 1. Zero public-ABI diff | pass — `abi-guard.sh check` clean, selftest PASS, `Include/` diff empty |
| 2. `pod lib lint` x3 podspecs | **iOS pass** ("passed validation"); tvOS not runnable locally |
| 3. Build iOS + tvOS | **iOS pass**; tvOS not runnable locally |
| 4. `trunk check` | pass — no new issues |
| 5. ObjC + Swift suites | pass — ObjC 979/979, Swift 327/327 (10 new here) |

tvOS remains verifiable only on CI: this machine has the tvOS 26.2 SDK but only
the 26.0 simulator runtime, so no tvOS destination resolves and pod lint's tvOS
leg fails the whole lint even though iOS passes. Environmental, not a compile
failure.

`UnitTests/SwiftTests/MParticle/MParticleSceneDelegateTests.swift` and
`UnitTests/Mocks/SceneDelegateHandlerMock.swift` are untouched and stay the
behaviour contract.


BrandonStalnaker (approved):
have some apprehension around naming (MPSceneDelegateLogic) but it could be changed when we make everything swift if we need too

PR #864, opened 2026-08-26
feat(migration): measure Swift progress against in-scope Objective-C

nickolas-dimitrakas:
## Why

The progress report divided Swift SLOC by **all** production Objective-C. That
denominator includes the surface we are deliberately keeping in Objective-C —
the public customer API, the kit and wrapper-SDK contracts, runtime-identity
pinned classes, and boundary glue (`CONVERSION-RECIPE.md` classifications 1, 3,
and 4). 100% was unreachable by construction, so the bar was measuring against
a target the migration is not pursuing.

## What changed

`Tools/swift-migration-retained-objc.txt` is a new reviewed manifest of the 34
Objective-C implementations that stay. The report removes them from the
percentage's denominator and reports their SLOC in a separate
`Objective-C retained` column. **100% now means every in-scope Objective-C
implementation is gone.**

### Effect on the current tree

| Area | Before | After | ObjC before | ObjC in scope | ObjC retained |
|---|---:|---:|---:|---:|---:|
| Core SDK | 24.50% | **34.65%** | 15,103 | 9,243 | 5,860 |
| SDK kit infrastructure | 0.00% | 0.00% | 3,725 | 3,340 | 385 |
| Standalone kits | 1.06% | 1.06% | 14,231 | 14,231 | 0 |

Every row reconciles exactly: in scope + retained = the previous remaining
figure. Kit infrastructure holds at 0.00% because it has no Swift yet.

### How the manifest was derived

Not by heuristic. Every entry backs a declaration in `Tools/abi-baseline.txt`
or is named in `CONVERSION-RECIPE.md`'s compatibility inventory:

- **Supported customer contracts** — `mParticle.m`, the `+MParticlePrivate`
  implementations of `MParticleOptions` / `MParticleSession` /
  `MPAttributionResult` / `MPNetworkOptions`, `MPEnums.m`'s exported constants,
  and the event, commerce, consent, identity, location, and Rokt declaration
  sites reached through `mParticle.h`.
- **Supported kit and wrapper-SDK contracts** — `MPKitAPI`, `MPKitExecStatus`,
  `MPKitRegister`, `MPSideloadedKit`, `MPForwardRecord`,
  `MPCommerceEventInstruction`, and the `Filtered*` types.
- **Runtime-identity pinned** — `MPUploadSettings` (classification 3).
- **Boundary glue** — `MPUserDefaultsConnector` (classification 4).

Deliberately **not** retained: the `_PRIVATE` accidental exports the recipe
lists as wrapper-removal candidates — `MPBackendController`,
`MPPersistenceController`, `MPNetworkCommunication`, `MPStateMachine`,
`MPNotificationController`, `MPKitContainer`, `SceneDelegateHandler`. Those are
~9k SLOC and stay in scope, which is why Core lands at 34.65% rather than
somewhere near done.

Standalone kits get no entries. Their `MPKit*` classes are resolved by name at
runtime, and no kit-side conversion has established that contract yet.

### Design decisions worth reviewing

- **Retaining a wrapper does not freeze it.** A contract wrapper keeps its
  `@interface`, class name, selectors, and nullability while its logic still
  moves to Swift. That thinning shows as a **falling retained figure**, not as
  percentage movement — so the ~5.9k retained SLOC stays visible instead of
  being written off. This is the main tradeoff: file-granular exclusion is the
  only thing `cloc` can express, so wrapper-internal progress lives in its own
  column rather than in the bar.
- **Both revisions are counted with the head revision's manifest**, so editing
  the manifest cannot book a redefinition as progress. This PR's own report
  reads `➖ 0.00 pp` on all three rows, which is the correct answer — no code
  moved.
- **Diff movement is left unfiltered.** Deleting lines from a retained wrapper
  is real work and still shows in the movement table.
- **Stale entries fail the report** rather than silently reverting a file to
  in-scope: a path missing at head, a non-`.m`/`.mm` path, or a path outside the
  three counted buckets all exit non-zero.

### Docs

- `PR-GATE.md` gains a **What "done" means** section, plus the requirement to
  keep the manifest in sync when classifying or deleting a retained wrapper.
- `CONVERSION-RECIPE.md` links classification 1/3/4 to the manifest.
- `Tools/README.md` documents the metric, the manifest format, the editing
  policy and the evidence an addition requires, and adds the manifest to the
  end-of-migration cleanup list.

## Gate results

| # | Item | Result |
|---|---|---|
| 1 | `Tools/abi-guard.sh check` | ✅ PASSED — no public ABI diff. Baseline untouched. `abi-guard-selftest.sh` also PASS. |
| 2 | `pod lib lint` on all 3 podspecs | ⚪️ Not run — diff touches only `Tools/` and `docs/`; no sources, headers, podspecs, or project files. Also blocked locally: this machine has the tvOS 26.2 SDK but only the 26.0 simulator runtime, so the tvOS leg cannot run here. CI is the sole tvOS verification. |
| 3 | Build green on iOS + tvOS | ⚪️ Not run — same reason; nothing compiled changed. Same tvOS runtime gap applies. |
| 4 | `trunk check` | ✅ Clean on all 5 changed files. `shellcheck -x` also clean. |
| 5 | ObjC + Swift unit suites | ⚪️ Not run — no SDK source changed. Tooling is covered instead by `swift-migration-progress.sh selftest`. |

**Tooling verification actually performed** (with `cloc` 2.10, SHA-256 verified
against the value pinned in `swift-migration-progress.yml`):

- `swift-migration-progress.sh selftest` → **PASS**. Fixture extended with a
  retained contract wrapper and a manifest; new coverage for denominator
  removal, retained thinning reported with a signed delta and zero percentage
  movement, comment/blank/whitespace parsing, and rejection of stale,
  non-Objective-C, out-of-bucket, and vendored entries.
- Real report generated on this tree — numbers in the table above.
- Stale-entry guard exercised against the real repository, not just the
  fixture: exits 1 with
  `retained manifest entry does not exist: <path>`.

No `CHANGELOG.md` entry — no public surface change.

### DIFF
--- a/Tools/README.md
+++ b/Tools/README.md
@@ -85,9 +85,36 @@ The self-test creates a temporary Git repository and verifies bucket
 isolation, exclusions, code/comment/blank handling, percentage rounding,
 zero-denominator handling, Objective-C++ counting, filenames with spaces,
 renames, physical diff movement across divergent histories, flat changes, and
-regressions.
+regressions. It also covers the retained manifest: comment/blank/whitespace
+parsing, removal from the percentage denominator, retained thinning reported
+without percentage movement, and rejection of stale, non-Objective-C,
+out-of-bucket, and vendored entries.
 
 ### Metric definitions
+
+Progress is measured against the Objective-C the migration actually intends to
+delete, so **100% means every in-scope Objective-C implementation is gone**, not
+that no Objective-C remains. The public/kit contract, runtime-identity, and
+boundary-glue implementations that stay Objective-C by design are listed in
+`Tools/swift-migration-retained-objc.txt`, removed from the percentage's
+denominator, and reported in their own `Objective-C retained` column.
+
+For each bucket:
+
+- **Progress** = Swift SLOC / (Swift SLOC + in-scope Objective-C SLOC).
+- **Objective-C in scope** = Objective-C SLOC still to be deleted.
+- **Objective-C retained** = Objective-C SLOC covered by the manifest, with a
+  signed delta when it moved.
+
+Retained wrappers keep their `@interface`, class name, selectors, and
+nullability, but their logic still moves to Swift and the wrapper thins to
+marshaling. That thinning shows up as a falling retained figure, not as
+percentage movement — the percentage tracks file removal, the retained column
+tracks the boundary shrinking.
+
+Both revisions are counted using the **head** revision's manifest. A PR that
+edits the manifest therefore re-measures its own base under the new definition
+instead of booking the redefinition as progress.
 
 The three production-code buckets do not overlap:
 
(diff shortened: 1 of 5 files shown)


nickolas-dimitrakas:
Superseded by #865: branch renamed to `ci/swift-migration-progress-goal-rows` for the semantic-branch check, title changed to `ci`, and the report now keeps the full-denominator number as a long-term goal row alongside the new in-scope short-term row.

PR #870, opened 2026-08-27
test: make workspace-switch and kit-registry concurrency tests deterministic

nickolas-dimitrakas:
> Stacked on #867. Review that one first — this PR's base is `ci/bound-simulator-and-network-hangs`, so the diff here is only the test changes.

## Why

These are the two tests that repeat most often in the ObjC crash failures blocking CI:

- `MParticleTests.testSwitchWorkspaceKitsWithStop`
- `MPKitContainerTests.testActiveKitsRegistryThreadSafety`

They turned out to have two completely different problems, and only one of them is a test problem.

## `testSwitchWorkspaceKitsWithStop` — genuinely a bad test

It waited on two nested hard-coded 10-second `dispatch_after` blocks, so it always burned **at least 20 seconds** of wall clock and depended on fixed timing rather than on the SDK actually being ready. Under CI load, "10 seconds is surely enough" is exactly the assumption that produces flakes.

It now waits on the SDK's own readiness signals — `MParticle.initialized` after `startWithOptions:`, then the replacement shared instance after `switchWorkspaceWithOptions:` — by spinning the main run loop until the condition holds, so main-queue work the SDK schedules still runs.

**The assertions are unchanged.** Same behaviour checked, no weakening.

| | Before | After |
|---|---|---|
| `testSwitchWorkspaceKitsWithStop` | 20.2s | **0.11s** |

Verified: 3/3 runs of the test alone, and 4/4 runs of the whole `MParticleTests` class (62 tests, 0 failures).

`MParticle.initialized` is exposed to the test via the existing `@interface MParticle ()` block, alongside the privates already redeclared there. No product change.

## `testActiveKitsRegistryThreadSafety` — the test is right, the SDK is wrong

This one is **not** flaky tooling. It is correctly catching a real data race, reproduced at **2 crashes in 10 runs of that test on its own**, nothing else running.

`flushSerializedKits` ([MPKitContainer.m:217](../blob/main/mParticle-Apple-SDK/Kits/MPKitContainer.m#L217)) fast-enumerates `kitsRegistry` and mutates it through `freeKit:` on the main queue **without taking `kitsSemaphore`** — the lock every other accessor of that set uses. `activeKitsRegistry` reads it under the semaphore; `configureKits:` mutates it under the semaphore. And `configureKits:nil` reaches the unlocked flush via an early return *before* the lock is acquired ([MPKitContainer.m:2027](../blob/main/mParticle-Apple-SDK/Kits/MPKitContainer.m#L2027)) — precisely what the test calls in a loop while three threads read the registry.

So the crash is mutation-during-enumeration on shared state, and the test is doing its job.

### What this PR does and does not do

Fixed here (test-side defects, real but not the cause of the crash):
- `encounteredError` was a plain `__block BOOL` written from four concurrent blocks — an unsynchronized flag is itself a data race. Now guarded by an `NSLock`, recording only the first exception.
- `XCTFail` was called from those background queues, which is not safe with XCTest. Failures are now reported from the `dispatch_group_notify` block on the main queue.

**Not fixed here:** the race itself. This test will still flake at roughly its current rate until `MPKitContainer` locking is addressed. That is deliberate — a core-concurrency change in a public SDK does not belong in a CI-stabilization PR, and AGENTS.md asks for coordination on anything touching the kit interface.

### Notes for whoever takes the race

Groundwork already checked:
- All three `flushSerializedKits` callers ([mParticle.m:611](../blob/main/mParticle-Apple-SDK/mParticle.m#L611), [mParticle.m:722](../blob/main/mParticle-Apple-SDK/mParticle.m#L722), [MPKitContainer.m:2029](../blob/main/mParticle-Apple-SDK/Kits/MPKitContainer.m#L2029)) are outside locked regions, so taking `kitsSemaphore` there would not deadlock.
- There is no `dispatch_sync` to the main queue inside any locked region.
- `freeKit:` must stay lock-free: `configureKits:` already calls it while holding the semaphore, and `dispatch_semaphore` is not recursive.
- The naive fix (wrap the main-queue flush in the semaphore) would let a long background critical section stall the main thread — the locked regions at 1149–1624 and 1656–1939 are large. Snapshotting the registry under the lock and iterating the snapshot off-lock avoids that, but `freeKit:` would still read `kitsRegistry` unlocked, so it is only a partial fix.

## Scope left alone

`testSwitchWorkspaceOptions` and the other tests in that family still use the 10-second `WORKSPACE_SWITCHING_DELAY` (three nested delays, ~30s). The macro is retained for them. The same treatment would cut roughly another 60s off the ObjC suite, but they were not among the tests flagged as flaky, so they are out of scope here.

No `CHANGELOG.md` entry: test-only, no consumer-facing SDK change.

### DIFF
--- a/UnitTests/ObjCTests/MPKitContainerTests.m
+++ b/UnitTests/ObjCTests/MPKitContainerTests.m
@@ -3438,12 +3438,33 @@
     dispatch_queue_t concurrentQueue = dispatch_queue_create("com.mparticle.test.concurrent", DISPATCH_QUEUE_CONCURRENT);
     
     NSInteger iterations = 100;
-    __block BOOL encounteredError = NO;
-    
+    // Shared across four concurrent blocks, so it is guarded rather than a plain
+    // __block BOOL - an unsynchronized flag is itself a data race, and reporting
+    // failures off the main thread is not safe with XCTest.
+    NSLock *errorLock = [[NSLock alloc] init];
+    __block NSException *firstException = nil;
+    __block NSString *firstExceptionSource = nil;
+
+    BOOL (^hasFailed)(void) = ^BOOL{
+        [errorLock lock];
+        BOOL failed = firstException != nil;
+        [errorLock unlock];
+        return failed;
+    };
+
+    void (^recordException)(NSException *, NSString *) = ^(NSException *exception, NSString *source) {
+        [errorLock lock];
+        if (firstException == nil) {
+            firstException = exception;
+            firstExceptionSource = source;
+        }
+        [errorLock unlock];
+    };
+
     // Multiple threads reading activeKitsRegistry
     for (NSInteger i = 0; i < 3; i++) {
         dispatch_group_async(group, concurrentQueue, ^{
-            for (NSInteger j = 0; j < iterations && !encounteredError; j++) {
+            for (NSInteger j = 0; j < iterations && !hasFailed(); j++) {
                 @try {
                     NSArray *activeKits = [kitContainer activeKitsRegistry];
                     // Access the returned array to ensure objects are valid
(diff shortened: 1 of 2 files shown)


BrandonStalnaker (approved):
This looks good but please make sure all the CI succeeds before merging

PR #890, opened 2026-09-01
fix: use supplied Rokt kit configuration

thomson-t:
## Background

Rokt already receives its configuration through the kit lifecycle. Reading the same data back through the core kit container creates an unnecessary dependency and can leave refreshed or switched workspace configuration out of sync.

## What Has Changed

- Use the configuration supplied to the Rokt kit as its source of truth.
- Read placement mappings and hashed-email settings directly from that configuration.
- Preserve later configuration updates through `setConfiguration:`.
- Clear stored configuration when the kit stops.
- Replace mocked-container coverage with launch, refresh, missing-configuration, and stop/relaunch tests.

## Screenshots/Video

N/A — no visual changes.

## Checklist

- [x] Self-review completed
- [ ] Tests added or updated
- [x] Tested locally

## Testing

- Trunk checks
- Rokt unit tests against the local core SDK
- iOS SDK build and unit tests

### DIFF
--- a/Kits/rokt/rokt/Sources/mParticle-Rokt/MPKitRokt.m
+++ b/Kits/rokt/rokt/Sources/mParticle-Rokt/MPKitRokt.m
@@ -5,9 +5,6 @@
 // Kit version
 static NSString * const kMPRoktKitVersion = @"9.4.1";
 
-// Constants for kit configuration keys
-static NSString * const kMPKitConfigurationIdKey = @"id";
-static NSString * const kMPRemoteConfigKitConfigurationKey = @"as";
 static NSString * const kMPAttributeMappingSourceKey = @"map";
 static NSString * const kMPAttributeMappingDestinationKey = @"value";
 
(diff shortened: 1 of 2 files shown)


BrandonStalnaker (dismissed):

nickolas-dimitrakas (dismissed):
Approving. Right fix, and it comes out net-negative on lines. Reading config back out of the container was a genuine hazard on workspace switch; using what the kit lifecycle already handed us removes the whole class of problem.

Confirmed the equivalence rather than assuming it: `MPKitConfiguration.m:56` sets `_configuration = configurationDictionary[@"as"]`, so `roktKit.configuration[key]` is exactly the old `kitConfig[@"as"][key]`. Lifecycle is coherent too — `roktKit` is `static __weak`, cleared in `stop` alongside `_configuration = nil`, and re-set on relaunch.

All four red checks are infrastructure, not this PR: both native-test jobs log `Restarting after unexpected exit, crash, or test timeout` in ObjC suites the Rokt kit isn't part of, and `Build onetrust` / `Build branchmetrics-3` died at 1h05 and 1h04 — timeouts.

Two follow-ups, neither blocking this:

1. **The core still has the old copy.** `mParticle-Apple-SDK/MPRokt.m:512` and `:546` still read `roktKitConfig[kMPRemoteConfigKitConfigurationKey][…]` through the container, so the stale-config behaviour this fixes in the kit is still present on the core-side path. #897 rewrites that file — worth sorting out between the two PRs which path is actually live for placements so the fix lands on both.
2. **Test isolation now leans on dealloc timing.** `testGetRoktHashedEmailUserIdentityTypeConfigNil` passes because `tearDown` drops the last strong ref and the weak `roktKit` follows. True today, but any future strong capture of the kit instance turns it into an order-dependent failure. An explicit `[self.kitInstance stop]` in `tearDown` is one line and makes it robust — this suite has enough flakiness already.

BrandonStalnaker (dismissed):

nickolas-dimitrakas (approved):

BrandonStalnaker (approved):

PR #894, opened 2026-09-01
refactor: evaluate event projections in Swift

thomson-t:
## Background

Projection matching and mapping are policy operations that can be evaluated in Swift without changing the SDK event objects delivered to integrations.

## What Has Changed

- Move event and commerce projection policy into a Foundation-only Swift engine.
- Preserve required, static, field, default, `forEach`, `last`, and custom-attribute behavior.
- Return projected descriptors and applied projection identifiers for Objective-C materialization.
- Add narrow per-kit launch-configuration lookup and reconfiguration operations.
- Stop internal callers from copying the container’s stored configuration directly.

## Screenshots/Video

N/A — no visual changes.

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally

## Testing

- Trunk checks and ABI guard
- Direct Swift projection-engine tests
- Objective-C projection and container regressions
- Local Rokt package tests


mParticle-Apple-SDK/Kits/MPKitContainer.m
```diff
@@ -859,489 +988,52 @@ - (void)project:(id<MPExtensionKitProtocol>)kitRegister
...
-                    [keys removeObjectsInRange:deletionRange];
-                }
-                
-                for (NSString *key in keys) {
-                    projectedDictionary[key] = userDictionary[key];
-                }
-            } else {
-                [projectedDictionary addEntriesFromDictionary:userDictionary];
+
+        if (output.projectionId) {
```
nickolas-dimitrakas: `osx.NumberObjectConversion`: converting `NSNumber * _Nullable` to a primitive boolean. The intent here is a nil check and the code does the right thing, so making it explicit clears the analyzer without changing behaviour.

```suggestion
        if (output.projectionId != nil) {
```

nickolas-dimitrakas (changes requested):
One analyzer fix, then I think this is good.

`build-and-lint / run-analyzer` is failing on a new clang warning — suggestion inline. Per `AGENTS.md` gotcha #1 the job fails on any surviving `: warning:` line; I pulled the full log and this is the only one.

Also worth dropping the CHANGELOG entry — per `AGENTS.md` those sections are generated from PR titles at release time and anything hand-written under `## [Unreleased]` never gets folded in.

Spot-checked the port and it holds up:
- `forEach` vs `last` is equivalent, including the catch-all `else` and the empty-collection case (`entities.last.map { [$0] } ?? []` vs the ObjC `numberOfProducts > 0` guard).
- The mirrored enums all match their ObjC raw values — `BehaviorSelector.forEach = 0`, `PropertyKind` 1-5, `CommerceKind` 1/2/3.

One note: 786 lines with 6 direct tests, and the `MPKitContainerTests.m` change is 94+/94- which reads as rename churn rather than new assertions. Worth confirming nothing got weakened in the rewrite.

nickolas-dimitrakas (dismissed):

nickolas-dimitrakas (approved):

BrandonStalnaker (approved):

PR #895, opened 2026-09-01
refactor: move kit container ownership to Swift

thomson-t:
## Background

`MPKitContainer_PRIVATE` is an internal runtime type whose Objective-C declarations were accidentally exposed as public SDK API. This completes the ownership cutover while retaining SDK-object execution at a private compatibility boundary.

## What Has Changed

- Make `MPKitContainer_PRIVATE` a Swift-owned runtime type.
- Move registry state and queued-call representation into the Swift module.
- Retain SDK object construction, marshalling, persistence access, and exception containment in a private Objective-C execution adapter.
- Move internal callers and test seams to private or opaque interfaces.
- Remove the Objective-C queue-item implementation.
- Remove the public container header, protocol, and `MParticle.kitContainer_PRIVATE` declaration.
- Update the audited ABI baseline and changelog.

## Screenshots/Video

N/A — no visual changes.

## Checklist

- [x] Self-review completed
- [ ] Tests added or updated
- [x] Tested locally

## Testing

- Trunk checks and ABI guard
- iOS SDK build
- All Objective-C test classes, including the dedicated 83-test container suite
- Swift unit suite, excluding the known unrelated date-formatter failure
- Local Rokt SPM tests
- Coverage: queue item 100%, Swift owner 83.8%, execution adapter 73.2%
- tvOS validation delegated to CI


nickolas-dimitrakas (changes requested):
Two things in the new Swift container.

**1. The registry lock can be bypassed.** Every other accessor goes through `registryLock`, but `kitsRegistry` hands back the mutable set itself — so anything holding it can change the set while another thread reads it, and the lock can't help at that point. The only users are two tests calling `removeAllObjects`. A small `resetRegistry()` that takes the lock would close the gap and only touch those two call sites. Worth doing while #878 is fixing a race in this same set.

**2. `preconditionFailure` takes the host app down.** `MPKitContainerExecutionAdapter` is only ever found by name at runtime — nothing references it directly, so in a static build the linker is free to drop it. Integration tests pass, so the paths we test are fine; but if it ever is dropped on a customer's setup, the result is a crash on SDK init rather than a degraded feature. Two cheap guards: give the adapter a `+load` so it's always linked in, and log-and-continue instead of trapping. We're a framework, so we shouldn't be the reason an app fails to start.

BrandonStalnaker (approved):

nickolas-dimitrakas (approved):

PR #897, opened 2026-09-01
feat: Migrate MPRokt to Swift

BrandonStalnaker:
## Background
- MPRokt is a partner-facing contract (Include/MPRokt.h, ABI-baselined, used by the public Swift overlay), so it stays ObjC. The .m mixed that API with Foundation-only work: kit-181 lookup, attribute remapping, placement-map JSON, sandbox defaulting, and hashed-email identity type.
- handleURLCallback: must return a real BOOL to the host app. Kits implement the selector without a Swift-importable protocol, so performSelector: is not enough; the ObjC code used NSInvocation.

## What Has Changed
- Extracted that logic into MPRoktLogicPRIVATE. MPRokt.m still owns public selectors, forwardSDKCall:, identity, user attributes, logging, RoktContracts types, and the message-queue vs main-queue split.
- Replaced NSInvocation with an IMP + unsafeBitCast to (AnyObject, Selector, NSURL) -> ObjCBool (NSInvocation is unavailable in Swift). Nil URL, missing kit, and unclaimed URL still return NO.
- Added MPRoktLogicTests; kept MPRoktTests.m. Public headers and selectors are unchanged.

## Checklist


nickolas-dimitrakas (approved):

mParticle-Apple-SDK-Swift/Sources/Rokt/MPRoktLogic.swift
```diff
@@ -0,0 +1,187 @@
...
+
+    // MARK: - Kit invocation
+
+    /// `performSelector:` cannot reliably return `BOOL`, and `NSInvocation` is
+    /// unavailable in Swift. Kits implement `handleURLCallback:` without a
+    /// Swift-importable protocol, so this calls the IMP directly and reads `ObjCBool`.
+    @objc(invokeHandleURLCallbackOnKit:url:)
+    public static func invokeHandleURLCallback(on kitInstance: Any?, url: URL?) -> Bool {
+        guard let url, let object = kitInstance as? NSObject else { return false }
+        let selector = NSSelectorFromString("handleURLCallback:")
```
thomson-t, line 142: Not in this PR, but as a next step, we can remove this mechanism and create a protocol, something like
```swift
@objc(MPRoktKitDispatchTarget)
public protocol MPRoktKitDispatchTarget {
    @objc(getSessionId)
    optional func getSessionId() -> String?

    @objc(handleURLCallback:)
    optional func handleURLCallback(_ url: URL) -> Bool

    @objc(logMParticleApiDiagnostic:)
    optional func logMParticleApiDiagnostic(_ code: String)
}
```
and then invoke the functions.

thomson-t (approved):

mParticle-Apple-SDK-Swift/Sources/Rokt/MPRoktLogic.swift
BrandonStalnaker: https://github.com/mParticle/mparticle-apple-sdk/pull/898

PR #900, opened 2026-09-01
fix: accept native GoDaddy R1 certificate chains on the 8.12.x line

jamesnrokt:
## Background
- 8.12.x pins the GoDaddy G2 intermediate and requires every certificate at chain index >= 1 to match, so it rejects any chain issued under GoDaddy's TLS Root CA R1 hierarchy. GoDaddy has ended DV issuance under G2, so affected hosts cannot be served a G2 leaf indefinitely.

## What Has Changed
- Backport root-only certificate matching from 15cf961cc (shipped in 8.26.0): match the last certificate in the evaluated chain instead of every intermediate.
- Replace the pinned G2 intermediate with the root set used by 8.26.0+, plus the self-signed GoDaddy TLS Root CA - R1.
- Add a regression test asserting the embedded R1 certificate matches GoDaddy's published SHA-256 fingerprint.
- Bump version metadata to 8.12.1.

## Screenshots/Video
- N/A — no visual changes.

## Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works.

## Additional Notes
- Retains the iOS/tvOS 9.0 deployment target, so `SecTrustGetCertificateAtIndex` / `SecTrustGetCertificateCount` are kept rather than the iOS 15+ replacements used on main. The added `certificateCount > 0` guard matters at this target: `SecCertificateCopyData(NULL)` crashes.
- `MPConnector.m` is byte-identical between v8.12.0 and v8.15.1, so this takes the v8.15.2 file verbatim — the diff here is character-for-character the v8.15.1..v8.15.2 diff for that file.
- CI on this branch pins retired runner images (macos-12, Xcode 13.4.1/14.1). Failures there are environmental, not caused by this change. Verified locally instead: iOS + tvOS build clean, and the iOS suite runs 568 tests with 3 failures — the same 3 that fail on unpatched v8.12.0 (locale and Foundation-internals assertions predating this change), so no new failures and no regressions.
- SPM: no artifacts or checksums to update on this line. 8.12.x `Package.swift` is source-based — `binaryTarget` first appears in 8.13.0 — so the tag alone carries the fix and the usual "update the URL and checksum together" step does not apply.
- Pre-built xcframeworks are attached to the v8.12.1 release for convenience (`ios-arm64`, `ios-arm64_x86_64-simulator`, `tvos-arm64`, `tvos-arm64_x86_64-simulator`), but `Package.swift` does not reference them and they are not required to consume this release. They are unsigned, matching what this line has always published.
- No Carthage entry added for 8.12.1: `mParticle_Apple_SDK.json` references fat-framework archives that this release does not produce, so 8.12.x Carthage users stay on 8.12.0.
- The added test asserts the pinned R1 certificate's fingerprint and fails if it is removed or altered; it does not exercise the challenge handler, so the root-selection change itself is covered by the local build and full-suite run rather than by a unit test. Note also that this widens trust relative to 8.12.0 — root-only matching, plus the ISRG roots that line did not previously trust — bringing 8.12.x in line with 8.26+. Pinning still runs after full `SecTrust` evaluation.


nickolas-dimitrakas (commented):
Is this meant to be pointing to the release branch? Or is it supposed to be the maintenance 8.x?

nickolas-dimitrakas (approved):

jamesnrokt:
> Is this meant to be pointing to the release branch? Or is it supposed to be the maintenance 8.x?

I branched 8.12.x from the 8.12.0 tag.

PR #911, opened 2026-09-01
refactor: move attribute validation to Swift

thomson-t:
## What

Second slice of the `MPBackendController.m` migration (stacked on #910). Extracts the key/value validation decision tree from `+[MPBackendController_PRIVATE checkAttribute:key:value:error:]` into a Foundation-only `MPAttributeValidator`, per `docs/swift-migration/CONVERSION-RECIPE.md`. Swift owns the decision and returns an `MPAttributeValidationResult`; the ObjC method keeps its selector, maps the result to the existing `NSError` codes, and keeps the console logging so the log level still gates message building.

## Contract preserved (covered by the existing `MPBackendControllerTests` checkAttribute cases + new `MPAttributeValidatorTests`)
- valid string / number / array-of-strings / `NSNull` values
- nil key and `NSNull` key → `kInvalidKey`
- key over `LIMIT_ATTR_KEY_LENGTH` → `kExceededAttributeKeyMaximumLength`
- nil value → `kNilAttributeValue` (no log — may be a removal)
- unsupported value type → `kInvalidDataType`
- string / array-combined length over `LIMIT_ATTR_VALUE_LENGTH` → `kExceededAttributeValueMaximumLength`
- array with a non-string entry → `kInvalidDataType`

## Deliberate note
The two **array-specific** console strings now reuse their scalar-equivalent error text. The return value and `NSError` code are unchanged, and the log stays gated behind the error log level exactly as before (verified by the `OCMock`-based `testCheckAttributeValueTooLong`, which asserts the value is not touched when logging is off).

## Gate
- `Include/` zero diff and `abi-guard.sh check` passes
- iOS build green; ObjC + Swift suites green; `trunk check` clean


nickolas-dimitrakas (dismissed):
Verified the migrated validation logic matches `checkAttribute:key:value:error:` exactly, including error codes and `MPIsNull` semantics. One non-blocking note: the array-specific console messages (per-entry value, combined-length text) now reuse the generic scalar wording, so a failed array attribute's log line no longer names the offending entry. Return values/error codes are unaffected — worth a quick follow-up for log detail, not a blocker.

nickolas-dimitrakas (approved):

PR #914, opened 2026-09-01
port: CI hang fixes and workspace-switch condition waits

nickolas-dimitrakas:
> **Draft.** Ports the subset of the main-line CI/test fixes that this branch is missing. Not a copy of #878 — see what was deliberately left out.

## Why

CI on PRs targeting this branch is failing and slow. From [run 33552140564](https://github.com/mParticle/mparticle-apple-sdk/actions/runs/33552140564) (PR #907):

| Job | Failure |
|---|---|
| `native-unit-tests (tvOS, mParticle-Apple-SDK)` | crash in `testSwitchWorkspaceKitsWithStop` |
| `native-unit-tests (iOS, mParticle-Apple-SDK)` | crash in `testActiveKitsRegistryThreadSafety` |

And the workspace tests are burning ~100 seconds of pure fixed delay per run:

```
testSwitchWorkspaceOptions          30.118s
testSwitchWorkspaceSideloadedKits   30.108s
testSwitchWorkspaceKitsWithoutStop  20.147s
testSwitchWorkspaceKitsNoConfigs    20.078s
```

None of the main-line fixes for this had reached this branch.

## What this ports

**CI workflows** — the #867 fixes: `-parallel-testing-enabled NO` (the Swift scheme was cloning simulators and starving the daemons one test blocks on), `simctl bootstatus`, a per-test timeout, plus the SPM resolve watchdog and job timeouts in `build-kits`. Taken from main after confirming this branch had no independent changes to those files.

**Five workspace-switch tests** — converted from nested 10-second `dispatch_after` chains to `MPWaitForCondition`, which spins the main run loop until the SDK is actually ready. `dispatch_after` and `WORKSPACE_SWITCHING_DELAY` are now gone from `MParticleTests`. Where a test asserts an outcome that arrives via the flush on the main queue, the wait includes that outcome so it cannot race the flush.

These were converted against **this branch's own test bodies**, not by copying main's file — that would have clobbered the migration-specific `resetRegistry`, `kitContainer_PRIVATE` and `registeredKits.anyObject` work here.

**`MPRoktTests` teardown** — stop the partial mock before releasing the object it wraps, rather than after.

## Deliberately not ported

**The `kitsRegistry` lock fix does not apply here.** The migration moved the registry into `MPKitContainer.swift`, where `registryLock` and `registry` are both `static` and every accessor takes the lock — `registeredKits` returns a locked copy, `removeRegisteredKit:` mutates under the lock. This branch's registry locking is sounder than main's; porting the fix would have been patching a defect that isn't here.

**The Rokt async timeouts and teardown drain are already present**, and `drainPendingAsyncWork` is a two-pass drain — more thorough than the main-line version.

## Honest limits

**The teardown ordering change is unvalidated.** `MPRoktTests` on this branch runs 10/10 clean locally *both with and without* it, so I have no local evidence it does anything. It is correct on its own terms — releasing a mocked object while its partial mock is installed is a real ordering bug — but I am not claiming it fixes anything observed.

**`testActiveKitsRegistryThreadSafety` is not addressed.** It crashes on this branch's CI too, and since the Swift registry here is correctly locked, the cause differs from the main-line one I fixed. Still open, still needs its own investigation.

**Local numbers are Xcode 26.3 / iOS 26; CI is Xcode 16.4 / iOS 18.** The workspace-test failures did not reproduce locally at all, so the conversion is justified by the CI evidence above plus the identical fix working on main, not by a local before/after.

### DIFF
--- a/.github/workflows/build-kits.yml
+++ b/.github/workflows/build-kits.yml
@@ -25,6 +25,7 @@ jobs:
     name: Pod Lint ${{ matrix.kit.name }}
     needs: load-matrix
     runs-on: macos-15
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
@@ -54,6 +55,7 @@ jobs:
     name: Build ${{ matrix.kit.name }}
     needs: load-matrix
     runs-on: macOS-latest
+    timeout-minutes: 60
     env:
       XCODE_VERSION: ${{ matrix.kit.xcode_version || '26.2' }}
       USE_LOCAL_VERSION: "1"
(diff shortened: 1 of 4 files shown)


nickolas-dimitrakas:
Superseded by #915 — `port/` is not an allowed branch prefix (semantic branch name check requires build/chore/ci/docs/feat/fix/perf/refactor/revert/style/test/). Rebranched as `fix/port-ci-and-concurrency-fixes-to-swift-migration`, same commits, same description.

PR #921, opened 2026-09-02
refactor: move message batching to Swift

thomson-t:
## Background

`MPBackendController`'s `batchMessageArraysFromMessageArray:` packs persisted messages into count- and byte-bounded uploads — pure arithmetic with no SDK-object, queue, or persistence dependency. Next Foundation-only slice extracted from the backend controller, continuing attribute validation (#911) and session timing (#912).

## What Has Changed

- Add `MPMessageBatcher.batchGroups(...)` (Foundation-only): packs message indices into batches by count and byte size, applies the larger crash-report ceilings, drops over-sized messages, returns index groups in order.
- Thin `batchMessageArraysFromMessageArray:` to marshal each `MPMessage` → `(byteLength, isCrashReport)`, call the Swift batcher, and rebuild `MPMessage` batches from the index groups. Crash byte limits stay in `MPIConstants.m`, passed in.
- Add Swift mirror tests (count / byte / per-message / crash-ceiling / mixed); keep the ObjC `MPBackendControllerTests` byte-limit suite as the behavior contract.

## Testing

- `trunk check` — clean
- `bash Tools/abi-guard.sh check` — no public ABI diff (signature unchanged)
- iOS build of `mParticle-Apple-SDK` — succeeded, 0 warnings
- Swift unit suite incl. new `MPMessageBatcher` mirror tests
- ObjC `MPBackendControllerTests` batch/byte-limit tests (9) — green
- Full ObjC suite, tvOS, three-podspec `pod lib lint` — delegated to CI

Continues the `MPBackendController` → Swift extraction; #911/#912 are already merged into `workstation/swift-migration`. Next in the stack (user-attribute value logic) will branch from this.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #922, opened 2026-09-02
refactor: move user-attribute value logic to Swift

thomson-t:
## Background

`MPBackendController` repeats three pure value transforms around user attributes: the storage null-sentinel ↔ `NSNull` normalization (both directions) and `NSDecimalNumber` increment arithmetic, duplicated across `userAttributesForUserId:`, `setUserAttributeChange:`, `incrementUserAttribute:`, and `incrementSessionAttribute:`. Continues the `MPBackendController` → Swift extraction.

## What Has Changed

- Add `MPUserAttributeLogic` (Foundation-only): `attributesFromStorage:` (sentinel string → `NSNull`), `attributesForStorage:` (`NSNull` → sentinel string), `incrementedValueFrom:byValue:` (`NSDecimalNumber` add on string values). The sentinel constant stays in `MPIConstants`, passed in.
- Delegate the four backend methods to the helper, removing the duplicated loops and inline decimal arithmetic. The mutation/decision core, opt-out check, `MPUserAttributeChange` construction, persistence, and logging stay in the ObjC boundary (value logging stays behind the log-level gate).
- Add Swift mirror tests (round-trip normalization; integer/negative/decimal increment); keep the ObjC `MPBackendControllerTests` attribute/increment suite as the behavior contract.

## Testing

- `trunk check` — clean
- Swift unit suite incl. new `MPUserAttributeLogic` mirror tests
- ObjC `MPBackendControllerTests` increment/attribute/session tests (14) — green

**Stacked on #921** (`refactor/swift-backend-batching`). Set the base to `workstation/swift-migration` in the UI once #921 merges.


nickolas-dimitrakas (dismissed):

LGTM.

nickolas-dimitrakas (approved):

LGTM.

PR #923, opened 2026-09-02
refactor: move user-identity read filtering to Swift

thomson-t:
## Background

`MPBackendController`'s identity read paths repeat pure array filtering: dropping malformed / out-of-range identity types, and removing the advertiser-id entry when ATT is not authorized. The identity-diff writer also converts a stored millisecond timestamp to a date. Continues the `MPBackendController` → Swift extraction.

## What Has Changed

- Add `MPUserIdentityLogic` (Foundation-only): `validIdentities:` (keep number types below the first non-user type), `identities:removingType:when:` (drop the advertiser-id entry when the caller flags ATT unauthorized), `dateFirstSetFromMilliseconds:` (ms → Date, or now).
- Delegate `identitiesForUserId:`, `userIdentitiesForUserId:`, and the `setUserIdentity:` first-set date computation to the helper. Identity-type / ATT thresholds stay in the ObjC enums, passed in.
- The `setUserIdentity:` **add/replace/remove/no-op decision core stays in the ObjC boundary** by design: it orchestrates the already-Swift `MPUserIdentityInstancePRIVATE` / `MPUserIdentityChangePRIVATE` objects, array mutation, and gated logging, with first-vs-last-match and empty-string/null subtleties that are safer left in place than marshalled for little gain.
- Add Swift mirror tests; keep the ObjC `MPBackendControllerTests` identity suite as the behavior contract.

## Testing

- `trunk check` — clean
- Swift unit suite incl. new `MPUserIdentityLogic` mirror tests
- ObjC `MPBackendControllerTests` identity tests (8) — green

**Stacked on #922** (`refactor/swift-backend-user-attribute-logic` ← `refactor/swift-backend-batching` ← workstation). Rebase the base down the chain in the UI as each merges.


nickolas-dimitrakas (dismissed):

LGTM.

nickolas-dimitrakas (dismissed):

LGTM.

nickolas-dimitrakas (approved):

PR #924, opened 2026-09-02
refactor: move crash/push message logic to Swift; drop dead backend state

thomson-t:
## Background

Two `MPBackendController` message paths carry pure computation: the crash-report base64 encoding with byte truncation, and the device-token change decision (no-op vs enabled/disabled status). The controller also declared two dispatch primitives that are never used. Final slice of the `MPBackendController` → Swift extraction wave (#921 / #922 / #923).

## What Has Changed

- Add `MPBackendMessageInfo` (Foundation-only): `base64CrashReport:maxBytes:` (UTF-8 → optional byte truncation → base64) and `pushRegistrationForDeviceToken:oldDeviceToken:` (returns the log token + `"true"`/`"false"` status, or nil when unchanged), plus the small `MPPushRegistrationDecision` result type.
- Delegate `logCrash:…` PL-crash encoding and `handleDeviceTokenNotification:` token decision to the helper. The caller keeps persistence, `MPMessageBuilder`, session selection, the message-data truncation, and logging.
- Remove dead code: `backendSemaphore` (created, never waited or signaled) and `backgroundSource` (declared, never used).
- Add Swift mirror tests; keep the ObjC `MPBackendControllerTests` crash and device-token tests as the behavior contract.

## Testing

- `trunk check` — clean
- Swift unit suite incl. new `MPBackendMessageInfo` mirror tests
- ObjC `MPBackendControllerTests` crash + device-token tests (8) — green

**Stacked on #923** (chain: #921 ← #922 ← #923 ← this). Rebase the base down the chain in the UI as each merges. Completes the current `MPBackendController` pure-logic extraction wave.


nickolas-dimitrakas (dismissed):

LGTM.

nickolas-dimitrakas (dismissed):

LGTM.

nickolas-dimitrakas (approved):

PR #929, opened 2026-09-02
refactor: move webview JSON field reading to Swift

nickolas-dimitrakas:
## Background

Continues the wrapper-deletion audit from #918: no further dead files or privatization candidates remain, so the series continues via extraction. `MPConvertJS.m` (webview→SDK-type JSON converter) was the best-shaped untouched candidate.

## What Has Changed

- Added `MPConvertJSFields.swift` reading `commerceEventActionForJSValue:`, `promotionFieldsFromJSON:`, `transactionAttributesFieldsFromJSON:`. SDK-type construction (`MPCommerceEvent`, `MPPromotion`, etc.) stays in the `.m`. **307 → 262 SLOC.**
- Preserved the JS-action↔SDK-action remap (orders differ), including returning `nil` for unknown/undefined values so the `.m` keeps its `MPILogError` + `AddToCart` fallback.
- `action` parameter typed `Any?` (not `NSNumber *`) since the original read a string-encoded value via `-intValue` with no enforced type.
- `.m` now assigns every field unconditionally — safe because every target setter already removes its key on nil.

## Validation

- `abi-guard` PASSED
- `trunk check` clean (caught 6 trailing-comma issues in new tests, fixed)
- ObjC 1004 / Swift 640 tests green; new Swift coverage 11/11
- tvOS: CI only

## Stack

Depends on #927.

## Checklist

- [x] Self-review completed
- [x] Tests added or updated
- [x] Tested locally


BrandonStalnaker (changes requested):
MPPromotion’s setters always go through the lazy attributes / beautifiedAttributes getters. Those allocate empty dictionaries, then remove a key that was never there. dictionaryRepresentation returns _attributes as-is (it does not treat empty as nil):

Before this PR, a promotion with no valid string fields left _attributes nil, so [promotion dictionaryRepresentation] was nil and MPPromotionContainer skipped it. After, it is @{}. The container then ships that empty object (and the same for beautified output):

nickolas-dimitrakas:
Good catch, @BrandonStalnaker — fixed. Restored the per-field `if` guard in `promotion:` so a nil value never touches the lazy `attributes`/`beautifiedAttributes` dictionary, keeping `dictionaryRepresentation` nil (not `@{}`) when nothing valid came through.

`transactionAttributes:` had the identical unconditional-assignment pattern and the identical lazy-dictionary setter, so it had the same bug — fixed that one too, same guard.

Added a regression test for each asserting `dictionaryRepresentation` is nil with no valid fields. Pushed as b8f17ca5.

BrandonStalnaker (approved):

PR #934, opened 2026-09-02
refactor: migrate network connector to Swift

thomson-t:
## Background

With request construction and URL pairing available in Swift, the remaining synchronous transport and certificate-pinning implementation can move without creating a module dependency cycle. This completes the connector migration while retaining the Objective-C runtime and protocol boundary used by the network coordinator.

## What Has Changed

- Moved connector transport, URLSession delegates, response handling, and certificate pinning to Swift.
- Preserved Objective-C runtime names, selectors, protocol conformance, HTTP status constants, and all six pinned roots.
- Added a Foundation-only connector configuration assembled by `MPNetworkCommunication`.
- Added Swift transport and request tests and adapted Objective-C integration tests to the generated interface.
- Removed the obsolete Objective-C connector implementation and private protocol headers.

## Validation

- `trunk check --ci`
- Public ABI guard and static analyzer
- Complete Objective-C and Swift iOS test schemes
- SwiftPM iOS and tvOS builds
- CocoaPods lint on iOS

The all-platform CocoaPods lint reaches and passes iOS, then cannot start tvOS validation because this machine has no tvOS simulator runtime.

## Stack

Depends on #933, which depends on #928.

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally


BrandonStalnaker (approved):
It'd be nice to have a more end-to-end test of MPNetworkCommunication to connector configuration to Swift connector assembly but we didn't have it before either

PR #936, opened 2026-09-02
refactor: move message dictionary construction to Swift

nickolas-dimitrakas:
## Background

`MPMessageBuilder.m`'s message-type↔string mapping, launch-info string, user-id filtering, and attribute-change/state-transition dictionary fragments are pure Foundation logic with no SDK-object dependency.

## What Has Changed

- Added `MPMessageBuilderFields.swift` (with a mirrored `MPMessageTypeSwift` enum, since the Swift module can't import the ObjC module).
- `MPMessageBuilder.m` keeps its class/selectors; only method bodies delegate to Swift. SDK-object marshaling (session/state-machine access, UIKit window-scene traversal, `MPMessage` construction) stays in ObjC.
- Preserved two quirks: `kMPMessageTypeStringPreAttribution` duplicates `Unknown`'s string value, so "unknown" never round-trips to `.preAttribution`; an out-of-range raw type still logs and falls back to `Unknown`.
- Added `MPMessageBuilderFieldsTests.swift`; existing ObjC coverage unchanged.

## Validation

- `abi-guard` PASSED, `trunk check` clean
- ObjC + Swift suites green
- tvOS: iOS-only locally (no simulator runtime on this machine)

## Stack

Depends on #932.

## Checklist

- [x] Self-review completed
- [x] Tests added or updated
- [x] Tested locally


BrandonStalnaker:
The Swift launchInfoString lookup used UIApplication.LaunchOptionsKey structs. ObjC callers would have passed the UIKit string constants, so both guards would fail and -launchInfo: would never write kMPLaunchURLKey. Bugbot caught that, and the new tests only constructed dictionaries with the Swift key type, so they wouldn’t have caught it either.

I don’t think we should fix the lookup. -launchInfo: has no callers in core, kits, or tests. App-state-transition launch data already goes through -stateTransition:previousSession:, which reads stateMachine.launchInfo (MPLaunchInfo), not a launch-options dictionary. The header isn’t in Include/ and isn’t in the ABI baseline.

Pushed 303fe906 to drop -launchInfo:, launchInfoStringFormat, the Swift helper, and those tests. Left -stateTransition: / MPLaunchInfo alone.

thomson-t (dismissed):

thomson-t (approved):

PR #940, opened 2026-09-03
feat: Migrate Rokt Kit + move logic into kit layer

BrandonStalnaker:
## Background
- Rokt-specific placement logic was split between the core SDK, the Objective-C Rokt Kit, and MPRoktLayout.
- Imperative placement attributes could bypass connection-level (hs.ua) filtering, while SwiftUI followed a separate preparation path.
- This duplication complicated the ongoing Swift migration and caused inconsistent mapping, identity, sandbox, and filtering behavior.

## What Has Changed
- Added a kit-owned Swift implementation target while retaining MPKitRokt as an ABI-stable Objective-C registration and forwarding shell.
- Moved Rokt lifecycle, SDK integration, identity synchronization, mapping, profile enrichment, session handling, diagnostics, and event handling into Swift.
- Reduced core MPRokt to public API marshaling, diagnostics, synchronous kit dispatch, timestamp creation, and queue ordering.
- Unified imperative and SwiftUI placement preparation.
- Placement attributes are now persisted, merged with profile attributes, and filtered through connection and data-plan rules before reaching Rokt.
- Added warnings for caller attributes blocked by filtering.
- Preserved post-filter injection of identities, mpid, and sandbox, including hashed-email precedence.
- Added MPID-scoped pending-write handling to preserve consecutive placement ordering without leaking attributes across users.
- Preserved Objective-C selectors, runtime registration, public headers, and SwiftPM/CocoaPods/Xcode packaging.
- Replaced migrated core and Objective-C implementation tests with Swift transformation tests and Objective-C boundary compatibility tests.
- Documented the major-release filtering impact. No privacy manifest changes are required.

## Checklist


Kits/rokt/rokt/Sources/mParticle-Rokt-Internal/MPRoktKitImplementation.swift
```diff
@@ -0,0 +1,961 @@
...
+/// Owns Rokt kit behavior while `MPKitRokt` retains the stable Objective-C runtime boundary.
+///
+/// `MPKitRokt.m` creates this object and forwards `MPKitProtocol` and
+/// `MPRoktKitDispatchTarget` selectors to it. Keeping registration and selector declarations in
+/// Objective-C preserves existing package and runtime behavior while the implementation lives here.
+@objc(MPRoktKitImplementation)
+public final class MPRoktKitImplementation: NSObject {
+    private enum Constants {
+        static let kitCode = 181
+        static let kitVersion = "9.4.1"
```
thomson-t, line 138: Can you move the version update logic https://github.com/mParticle/mparticle-apple-sdk/blob/main/.github/workflows/release-draft.yml#L92 to here?

BrandonStalnaker: good catch. I updated the logic to update this line rather than the old one.

thomson-t (approved):

PR #942, opened 2026-09-03
refactor: remove the orphaned MPAliasResponse

thomson-t:
Stacked on #941 — base is `refactor/migrate-response-value-types-to-swift`, so review that one first.

## Summary

Removes `MPAliasResponse`. It is a public class the SDK can no longer deliver to anybody.

- Deletes `Include/MPAliasResponse.h` and `Identity/MPAliasResponse.m`.
- Deletes the discarded `MPAliasResponse` construction from `-performAliasUpload:`, along with the
  re-parse of the just-uploaded request body that existed only to populate it.
- Replaces `MPAliasResponsePlanPRIVATE` (added in #941) with `MPAliasUploadOutcomePRIVATE`, trimmed
  to the three classifications that actually drive the upload loop — `isSuccessCode`,
  `isInvalidCode`, `shouldRetry` — while keeping the server failure logging.
- Drops the gratuitous `#import "MPAliasResponse.h"` from `Include/MPIdentityApi.h`; that header
  never named the type.
- Updates `Tools/abi-baseline.txt`, `Tools/swift-migration-retained-objc.txt`, and `MIGRATING.md`.

Net 434 deletions against 178 insertions.

## Audit: why this declaration may leave the public surface

Per `PR-GATE.md` item 1, a baseline update requires the compatibility audit recorded here.

**It had a delivery path, and that path was deliberately removed.** `MPAliasResponse` was
introduced in `e8516877` together with `-onAliasRequestFinished:` on `MPListenerController`, a
public callback declared in `Include/MPListenerController.h`. PR #495 (`b59cfc7e`, 2026-01-21)
deleted `Include/MPListenerController.h` and `Include/MPListenerProtocol.h` and documented it at
`MIGRATING.md` → *"Removed MPListenerController — The SDK no longer invokes any listener
callbacks."* `MPAliasResponse` was left behind.

**Nothing has surfaced one since.** No callback, notification, return value, or typedef carries the
type. There is no `MPAliasCompletionHandler`. `-[MPIdentityApi aliasUsers:]` returns `BOOL`. So no
application can observe an instance the SDK produced, and the object built at
`MPNetworkCommunication.m:892` was discarded on every alias upload.

**Usage audit, all clear:**

| Surface | Result |
| --- | --- |
| Kits (`Kits/`, `mParticle-Apple-SDK/Kits/`) | no references; `MPKitProtocol.h` has no alias selector |
| Wrapper-SDK integration points | no references |
| Swift umbrella (`MParticle/Sources/`) | no references |
| Sample and integration apps (`Example/`, `SwiftExample/`, `IntegrationTests/`, `RNExample/`) | no references |
| Archiving / `NSSecureCoding` / `NSCopying` | none — no conformances |
| `NSClassFromString`, KVC, nib/storyboard | none |
| Remaining repo references after this PR | `MIGRATING.md` only |

**Precedent for the release vehicle.** `MIGRATING.md` records `MPListenerController`,
`AppDelegateProxy`, the deprecated `UIApplicationDelegate` methods, and Location support all being
removed in **9.0.x → 9.1.0**, so this repo removes public declarations in minor releases with a
migration note rather than holding them for a major.

**Baseline diff is exactly the audited symbols** — 7 lines, one `IFACE` and six `PROP`
(`responseCode`, `errorResponse`, `request`, `requestID`, `willRetry`, `isSuccessful`), nothing
else. Regenerated with `Tools/abi-guard.sh update` from the final tree; `check` then exits 0.

## Nothing is lost if an alias callback returns

If alias-result reporting comes back — parity with the Android and Web SDKs — the information is
still available. #941 moved the full response decode into Swift and it stays unit-tested; only the
fields with no consumer were dropped here. Re-adding a callback means vending from the outcome type
rather than re-deriving anything, and the wire keys already live in `IdentityHTTPKeys`.

**If a callback is already on the roadmap, say so and I'll restore the DTO instead** — this PR is
the right call only if it is not.

## Behaviour

Unchanged. The removed object had no observers; the retained classification is byte-for-byte the
same set of status-code decisions, and the "Alias request failed - `<code>` `<message>`" error log
is preserved. The three ObjC tests in the deleted `MPAliasResponseTests.m` covered only
`isSuccessful` and property storage; `MPAliasUploadOutcomeTests` covers the same 2xx boundary
directly, plus the 4xx/429/503/5xx classifications and four malformed-body cases the ObjC tests
never reached.

## Open question for the reviewer

`MIGRATING.md` gets a new `## Migrating from 9.4.x to 9.5.0` heading. `VERSION` is 9.4.1, so
correct the heading if this lands under a different number — the release workflow does not rewrite
`MIGRATING.md`.

## Validation

| Gate | Result |
| --- | --- |
| `bash Tools/abi-guard.sh check` | PASSED against the regenerated baseline |
| `trunk check --ci` | no new issues |
| iOS framework build | BUILD SUCCEEDED |
| `xcodebuild analyze` (iOS) | no `: warning:` lines |
| Objective-C suite (iPhone 16 Pro) | TEST SUCCEEDED |
| Swift suite (iPhone 16 Pro) | TEST SUCCEEDED |
| `pod lib lint` all three podspecs, iOS | passed validation, no warnings |
| SwiftPM graph, iOS | Build complete |
| tvOS build / tvOS pod lint | **not run — tvOS 26.5 is not installed in this environment** |

Same tvOS limitation recorded on #926, #928, #933, #934 and #941. One unrelated test
(`MPKitContainerTests testActiveKitsRegistryThreadSafety`) failed once under parallel execution and
passed on re-run; it is a thread-safety test and touches nothing here.

## Checklist

- [ ] Tests added or updated
- [ ] Documentation updated
- [ ] Breaking change documented in `MIGRATING.md`


thomson-t:
Closing per product decision: `MPAliasResponse` stays. We do not want a breaking change or a `MIGRATING.md` entry for it, so the class keeps its public Objective-C declaration.

#941 is unaffected — it never touched `Include/MPAliasResponse.h` or `Tools/abi-baseline.txt`, and it keeps `MPAliasResponse.m` as a thin wrapper. The audit recorded above is left here for reference if the question comes up again.

PR #943, opened 2026-09-03
refactor: route network upload deletion through a Swift persistence seam

thomson-t:
## Background

`MPNetworkCommunication` reaches the ObjC `MPPersistenceController` directly for the only persistence op the upload paths use: `deleteUpload:` (×4, in `performMessageUpload:`/`performAliasUpload:`). To let upload handling move to Swift next (Swift can't import the ObjC module), that dependency needs a Foundation-only seam. Continues the network migration (transport already Swift — #928/#933/#934; response mapping in #941).

## What Has Changed

- Add `MPUploadPersisting` — Foundation-only `@objc` protocol, `deleteUpload(_: MPUploadPRIVATE)` — the seam a future Swift upload handler will take instead of the concrete controller.
- Add a thin private `MPUploadPersistenceAdapter` in `MPNetworkCommunication.m` conforming to it and forwarding to `MParticle.sharedInstance.persistenceController` (boundary glue, in-file → no new ObjC file / pbxproj churn).
- Route the four upload-path `deleteUpload:` calls through an injectable `id<MPUploadPersisting> persistence` (lazily the adapter). **No behavior change** — NC now depends on the Swift protocol, not the concrete controller. Consumer-info ops stay direct (they pass the ObjC `MPConsumerInfo`, only in the ObjC `parseConfiguration`) — deferred.

## Testing

- `trunk check` + `abi-guard.sh check` — no public ABI diff (NC signatures unchanged)
- iOS build — 0 warnings
- ObjC `MPNetworkCommunicationTests` upload/delete tests (7) green — seam exercised end-to-end
- Full ObjC suite / tvOS / pod lib lint → CI

No Swift mirror test: this is a plumbing seam (no logic), so the ObjC upload tests are the contract.

**Stacked on #941.** Retarget to `workstation/swift-migration` once #941 merges. **Next:** an `@objc` MParticle-access context → move the upload effect handling (delete/logKitBatch/throttle) into a Swift upload handler consuming this seam.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #944, opened 2026-09-03
refactor: move message-upload response handling to Swift

thomson-t:
## Background

After the transport + response-mapping migrations, `performMessageUpload:` still classified the HTTP response and applied its effects inline in ObjC. This moves that decision + the Swift-reachable effects into a Swift handler, consuming the persistence seam (#943) and a new kit-batch-logging seam. Continues MPNetworkCommunication → Swift.

## What Has Changed

- Add `MPUploadResponseHandler` (Foundation-only): classifies the response (success 200–299 / invalid `!=429 && 400–499`), resets the transport-error counter, deletes the upload (persistence seam) and logs the kit batch (new `MPKitBatchLogging` seam) on success, and returns an `MPUploadResponseOutcome` (`isSuccess`/`willRetry`/`shouldThrottle`/`throttleRetryAfter`).
- Add a thin private `MPKitBatchLoggingAdapter` in `MPNetworkCommunication.m` forwarding to `MParticle.logKitBatch:`.
- Rewire `performMessageUpload:`'s response tail to call the handler, then apply the state-machine throttle + the success-body config parse (kept in ObjC — touches persistence) from the outcome. **No behavior change.**

## Testing

- `trunk check` + `abi-guard.sh check` — no public ABI diff (NC signatures unchanged)
- iOS build — clean
- Swift unit suite incl. new `MPUploadResponseHandler` mirror tests (success / invalid / 429 / 503 / 5xx / transport-retriable)
- ObjC `MPNetworkCommunicationTests` upload tests (7) green — success/invalid deletion, ATT, offline, zip-fail identical

**Stacked on #943** (chain: #941 ← #943 ← this). Retarget down the chain in the UI as each merges. **Next:** apply the same handler to `performAliasUpload:` (dedup with #941's alias plan) and, eventually, the ATT device-dict transform.


BrandonStalnaker (dismissed):
Looks good in general

Nit: The MPILogWarning “Throttling uploads after transport error.” is gone from the message path (alias still logs). You might want to restore it in the Swift handler or next to throttleWithRetryAfter:.

BrandonStalnaker (approved):

PR #945, opened 2026-09-03
refactor: move upload ATT device-info transform to Swift

thomson-t:
## Background

`performMessageUpload:` mutated the event upload's device-info block inline to apply the App Tracking Transparency status (wire status string, advertiser-id removal, ATT timestamp) — a pure JSON transform, the last self-contained message-upload slice still in ObjC.

## What Has Changed

- Add `MPUploadDeviceInfo.uploadDataApplyingATT(_:authStatus:authTimestamp:)` (Foundation-only): parses the upload JSON, writes the ATT wire status, drops the advertiser id for every status except authorized, stamps the ATT timestamp, re-serializes. Returns the input unchanged if it is not a JSON object. Wire keys mirror MPIConstants (`di`/`atts`/`attt`/`aid`).
- Replace the inline transform in `performMessageUpload:` with one call; ObjC keeps resolving the ATT status/timestamp from the state machine and owns compression/IO.
- Add Swift mirror tests for each ATT status + non-JSON input; keep the ObjC ATT upload tests as the behavior contract.

## Testing

- `trunk check` + `abi-guard.sh check` — no public ABI diff
- iOS build — clean
- Swift unit suite incl. new `MPUploadDeviceInfo` mirror tests
- ObjC `MPNetworkCommunicationTests` ATT tests (Authorized/Denied/NotDetermined/Restricted) + deletion/zip-fail (6) green
- Note: the Swift `MPUserDefaultsTests.testSaveRestore` failure in the full-suite run is a **pre-existing order dependency** (passes in isolation), unrelated to this change.

**Stacked on #944** (chain: #941 ← #943 ← #944 ← this). Retarget down the chain in the UI as each merges.


BrandonStalnaker (changes requested):
For the old Objective-C, if di is missing, it messages nil and never writes a device block. New Swift does `(uploadDict["di"] as? [AnyHashable: Any]) ?? [:]`, then always stamps attt (and atts / drops aid when unauthorized) onto a new di.

That can add a device-info object (and ATT metadata) on payloads that previously shipped without one. Either return the original original funcitonality when di is absent or not a dictionary, or treat synthesis as an intentional fix and add a test for the missing-di case.

thomson-t:
> For the old Objective-C, if di is missing, it messages nil and never writes a device block. New Swift does `(uploadDict["di"] as? [AnyHashable: Any]) ?? [:]`, then always stamps attt (and atts / drops aid when unauthorized) onto a new di.
> 
> That can add a device-info object (and ATT metadata) on payloads that previously shipped without one. Either return the original original funcitonality when di is absent or not a dictionary, or treat synthesis as an intentional fix and add a test for the missing-di case.

Thanks. Updated in db16f6ae

BrandonStalnaker (approved):
Thanks for the change

PR #947, opened 2026-09-03
refactor: move config Cache-Control max-age parsing to Swift

thomson-t:
## Background

`requestConfig:` parses the config response's `Cache-Control` header for the `max-age` seconds (clamped to the max expiration age) via `maxAgeForCache:` — a pure string parse, the cleanest slice of that otherwise boundary-heavy method (user-defaults, connector, `MPResponseConfig`, background tasks stay ObjC).

## What Has Changed

- Add `MPConfigResponseParser.maxAge(fromCacheControl:maxExpirationAge:)` (Foundation-only): case-insensitive `max-age=` extraction, `NSString.doubleValue` leading-number parse, clamped to the limit; nil when absent.
- `maxAgeForCache:` becomes a one-line wrapper delegating to it (kept because the ObjC test suite calls it directly and `requestConfig:` uses it). The limit stays in `MPIConstants`, passed in.
- Add Swift mirror tests; keep the ObjC `testMaxAgeForCache*` suite as the behavior contract.

## Testing

- `trunk check` + `abi-guard.sh check` — no public ABI diff
- iOS build — clean
- Swift unit suite incl. new `MPConfigResponseParser` mirror tests
- ObjC `MPNetworkCommunicationTests` `testMaxAgeForCache*` + complex-cache-control (7) green

**Stacked on #945** (chain: #941 ← #943 ← #944 ← #945 ← this). Retarget down the chain in the UI as each merges.


BrandonStalnaker (dismissed):
Looks exactly the same as the old string parse

BrandonStalnaker (approved):

PR #952, opened 2026-09-04
refactor: dissolve the abstract data-model base class

thomson-t:
Moves two data-model types to Swift, removes the abstract base class they only inherited a `uuid`
property from, and records two Objective-C files as intentionally retained.

## 1. Remove `MPDataModelAbstract`

`MPDataModelAbstract` provided its subclasses with a `uuid` property and an `NSCopying`
implementation and nothing else. Its three subclasses were already thin Objective-C shells
forwarding to Swift implementations, so the base class was the only thing preventing two of them
from being removed outright.

- `MPBreadcrumbPRIVATE` now carries the `MPBreadcrumb` Objective-C runtime name and absorbs the
  behaviour the shell still held: `NSCopying`, `NSSecureCoding`, `description`, and an
  `isEqual(_:)` override that type-checks its argument as the previous null guard did.
- `MPIntegrationAttributesPRIVATE` now carries the `MPIntegrationAttributes` runtime name. Its
  shell was pure forwarding and nothing read `uuid` or copied it, so no other work was required.
- Removed: `MPBreadcrumb.{h,m}`, `MPIntegrationAttributes.{h,m}`, `MPDataModelAbstract.{h,m}`,
  `MPDataModelProtocol.h`, and `Sources/Utils/MPDataModelAbstract.swift` with its test — the
  Swift helper's only caller was the removed `MPDataModelAbstract.m`.
- `MParticleUserNotification` keeps its Objective-C shell and declares its own `uuid` property in
  place of the inherited one.

`MPDataModelProtocol` is removed with them. Nothing anywhere performs a `conformsToProtocol:`
check against it, and its only two adopters were the headers being deleted.

### Public API and ABI are unaffected

Neither type has a header under `mParticle-Apple-SDK/Include/`, and neither carries
`ATTRIBUTES = (Public, )` in the Xcode project. Their names do appear in exported headers, as
forward declarations and in method signatures recorded in `Tools/abi-baseline.txt`:

- `Include/MPPersistenceController.h` — `@class MPBreadcrumb;`, `-fetchBreadcrumbs`
- `Include/MPPersistenceController.h` — `@class MPIntegrationAttributes;`,
  `-saveIntegrationAttributes:`, `-fetchIntegrationAttributes`

`Tools/abi-guard.sh` reads class, protocol, property, method and enum declarations from
`Include/*.h`; it does not capture forward declarations or typedefs. Because the Swift types keep
the same Objective-C runtime names, every recorded signature stays byte-identical. `abi-guard.sh
check` exits 0 and the baseline is unchanged.

### Behaviour kept as-is

Two existing quirks are load-bearing and are preserved deliberately, with comments in the source
explaining why:

1. Breadcrumb equality returns `false` when both sides have a nil `uuid`, `sessionUUID` and
   `breadcrumbData`. An existing test asserts this.
2. Encoding persists `content` while decoding rebuilds `breadcrumbData` from it as UTF-8. Archive
   round-trip equality depends on that asymmetry.

### Test target note

`UnitTests/Mocks/MPPersistenceControllerMock.swift` needed
`internal import mParticle_Apple_SDK_Swift` and now names `MPIntegrationAttributesPRIVATE`. It
previously reached the type through the bridging header, and importing the umbrella module
re-exports only the Objective-C module. In Swift the symbol is the `PRIVATE` name;
`MPIntegrationAttributes` is only the Objective-C runtime name.

## 2. Two files recorded as intentionally retained

`Tools/swift-migration-retained-objc.txt` lists the Objective-C implementations that are not
expected to be removed, and separates the two progress figures the migration reports. Both files
below were counted as still-to-be-migrated despite having no path to Swift, which makes the
in-scope figure unreachable. Adding them corrects that denominator.

**`MPIConstants.m`** — 360 C global definitions (`NSString *const`, `const NSTimeInterval`,
`const NSInteger`) declared `extern` in an internal header and referenced by bare identifier from
51 Objective-C files. Swift cannot emit C globals, so these symbols cannot move; the Swift module
mirrors the values it needs privately, as in `Sources/MessageKeys.swift`. The file is also written
to by release tooling, which rewrites `kMParticleSDKVersion` in it.

**`MPDataPlanFilter.m`** — its entire interface is SDK types the Swift module cannot import:

```objc
- (instancetype)initWithDataPlanOptions:(MPDataPlanOptions *)dataPlanOptions;
- (MPBaseEvent *)transformEventForBaseEvent:(MPBaseEvent *)baseEvent;
- (MPEvent *)transformEventForEvent:(MPEvent *)mpEvent;
- (MPCommerceEvent *)transformEventForCommerceEvent:(MPCommerceEvent *)commerceEvent;
```

`MPEvent` and `MPCommerceEvent` are also fixed by the kit interface, and `transformEventForEvent:`
mutates the event in place. The portable logic has already moved: the decision helpers live in
`Sources/Kits/MPDataPlanFilterPolicy.swift`, and `MPDataPlanFilterProtocol` is the Swift-facing
boundary. What remains is the marshalling that boundary requires.

### Audit for both entries

| | `MPIConstants.m` | `MPDataPlanFilter.m` |
| --- | --- | --- |
| `Tools/abi-baseline.txt` | no entries — header not under `Include/` | no entries — header not under `Include/` |
| Exported header | no | no |
| Referenced under `Kits/` | none | none |
| Wrapper-SDK integration points | none | none |
| Archiving, dynamic lookup, KVC | n/a — constants only | none — no coding conformance |

`Tools/README.md` asks for the corresponding `Tools/abi-baseline.txt` entry when adding to this
list. Neither file has one, because neither header is exported — the same is true of the entry
already listed for `MPUserDefaultsConnector.m`.

## Effect on the reported figures

| | Base | This PR |
| --- | ---: | ---: |
| Core SDK, in-scope Objective-C | 6,711 | **6,571** |
| Core SDK, in-scope progress | 57.09% | **57.70%** |
| Kit infrastructure, in-scope Objective-C | 734 | 734 |

The `+0.61 pp` comes from the code removed in part 1. The two list entries contribute nothing to
the reported change by design — the report measures both revisions using the list from the head
revision, so an entry cannot inflate a delta. Their effect is on the absolute figures from here on.

## Testing

Verified locally on iOS: framework build, static analyzer with no new warnings, the full
Objective-C and Swift unit-test suites, and `pod lib lint` across all three podspecs. The
Objective-C contract tests covering this change pass unmodified — 80 tests across
`MPDataModelTests`, `MPIntegrationAttributesTest` and `MPPersistenceControllerTests`, including
breadcrumb copy equality, the null and wrong-type equality cases, and the archive round trip.
`abi-guard.sh check` exits 0 with no baseline change, and the migration report's self-test passes
with both new list entries validating.

tvOS was not exercised locally; it is covered by CI.

## Not included

- `MPPersistenceController.m` is untouched. Its status is worth settling separately: it is
  described elsewhere as staying Objective-C because of its direct SQLite use, while
  `docs/swift-migration/CONVERSION-RECIPE.md` lists `MPPersistenceController_PRIVATE` among the
  types expected to be removed. At 1,787 lines the answer materially changes what the in-scope
  figure means.
- `MPNotificationController.m` is untouched and is not a candidate for this list.
  `docs/swift-migration/CONVERSION-RECIPE.md` expects it to be removed eventually. It is blocked
  today because `-setDeviceToken:` reaches the network layer through `MParticle`, but that calls
  for a protocol boundary rather than a reclassification.
- The projection types (`MPBaseProjection`, `MPEventProjection`, `MPAttributeProjection`,
  `MPProjectionMatch`) are movable and left for a focused change. They share protected instance
  variables across a two-level hierarchy, implement polymorphic copying that writes into those
  variables, and use failable initialisers that propagate `nil` through the chain, so all four have
  to move together.

## Checklist

- [ ] Code change reviewed
- [ ] Both retained-list entries reviewed and agreed
- [ ] Public API unchanged


thomson-t:
Rebased onto `63a6e774` to clear the conflict after #940, #951, #953, #956, #958, #959 and #963
merged. Force-pushed; the three commits are unchanged in content.

Two conflicts, both collisions in test-file import blocks with #951:

- `UnitTests/ObjCTests/MPDataModelTests.m` — #951 replaced `#import "MPMessageBuilder.h"` with
  `#import "MPIConstants.h"` at the same position this branch removes `#import "MPBreadcrumb.h"`.
  Resolved by keeping `MPIConstants.h` and removing `MPBreadcrumb.h`.
- `UnitTests/ObjCTests/MPPersistenceControllerTests.m` — both sides removed an adjacent import
  (#951 removed `MPMessageBuilder.h`, this branch removes `MPIntegrationAttributes.h`). Resolved by
  taking both removals.

`project.pbxproj` merged automatically against the entries added by #951 and #963 with no manual
intervention, and both `Tools/swift-migration-retained-objc.txt` commits applied cleanly.

Re-verified after the rebase: `abi-guard.sh check` exits 0 against the updated baseline (this
branch still changes none of it), iOS build succeeds, the static analyzer reports no new warnings,
and the full Objective-C and Swift suites pass along with the 80 contract tests covering this
change.

Unrelated, for whoever picks it up: the set of Swift test files that fail `trunk fmt` on
`workstation/swift-migration` has grown from 11 to 14 — `MPAttributeValidatorTests.swift`,
`MPBackendMessageInfoTests.swift` and `MPUserAttributeLogicTests.swift` came in with #956 and #959.
None is part of this diff.

BrandonStalnaker (approved):

PR #953, opened 2026-09-04
refactor: move upload batch preparation logic to Swift

thomson-t:
Stacked on #951 (`refactor/swift-message-builder`).

Two pure slices of the upload-preparation path move to `MPUploadBuilderFields`, which already owns every other pure computation on this path.

## What changed

**`MPUploadGrouping.groupsFromStoredMessages:`** flattens the four-deep `mpid → sessionId → dataPlanId → dataPlanVersion` nesting that `fetchMessagesForUploading` returns into one `MPUploadMessageGroup` per innermost bucket, decoding the persistence sentinels (`sessionId == -1`, `dataPlanId == "0"`, `dataPlanVersion == 0`) to nil on the way. `MPMessage` is already a Swift type, so nothing is marshalled.

`prepareBatchesForUpload:` loses its four nested enumeration blocks and becomes a flat loop. It still owns the fetch, the `MPUploadBuilder` loop, `saveUploads:deleteMessages:` and `deleteAllSessionsExcept:`.

**`MPUploadBuilderFields.preparedMessagesFrom:`** takes over the per-message enumeration in `MPUploadBuilder`'s initializer: opt-out detection, prepared message ids, message dictionaries. The ids and the dictionaries are deliberately not parallel — every message contributes an id so it is deleted after upload, but only those that serialize contribute a dictionary, which is what the original did.

Both helpers skip entries whose shape does not match instead of crashing, identical for every well-formed payload. Empty buckets still produce a group, because the caller runs its save/delete per group.

## Why `MPUploadBuilder` is not deleted here

It looked like a twin of `MPMessageBuilder` — internal header, one production caller, no archive identity — but its `build:` reaches the persistence controller (5 methods), the identity API, `MParticleOptions.onCreateBatch`, `MPConsumerInfo` and `MPConsentState`, and **every pure computation it performs already delegates to `MPUploadBuilderFields`**. What is left is marshalling, which `CONVERSION-RECIPE.md` explicitly keeps at the Objective-C boundary:

> If a conversion seems to need the Swift side to know about an SDK type, that's a sign the boundary is drawn wrong — push the marshaling back into the `.m`.

Making it a Swift class would need roughly five new seams whose Objective-C adapters would perform the same marshalling the `.m` does today — moving lines rather than removing them, and improving the migration percentage without doing real work. So this PR takes the genuinely pure slices and leaves the class where it is.

## Validation

- `Tools/abi-guard.sh check` → exit 0, no baseline or retained-manifest change
- `trunk check` → clean
- iOS build → 0 errors
- Swift suite: **831 passed, 0 failed**, including 12 new mirror tests (`MPUploadGroupingTests`, `MPPreparedMessagesTests`)
- Objective-C contract tests: **177/177** across `MPUploadBuilderTests`, `MPBackendControllerTests`, `MPPersistenceControllerTests`, `MPMessageBuilderTests`, `MPDataModelTests` — run twice, clean both times


BrandonStalnaker (dismissed):
Sentinel decoding is the same (sessionId == -1, dataPlanId == "0", version == 0). Every message still contributes an id; only serializable messages contribute a dictionary; NSNull is skipped. Empty persistence still skips the loop and does not clear deletedUserAttributes.

BrandonStalnaker (approved):

PR #956, opened 2026-09-04
refactor: move backend upload-trigger and clean-up decisions to Swift

thomson-t:
Stacked on #953 (`refactor/swift-upload-batch-preparation`).

Three pure decisions in `MPBackendController` move into the Swift helpers that already own their neighbours. No new files — each slice joins the helper it belongs with.

## What moved

**`MPBackendMessageInfo.shouldUploadMessageOfType:...`** — the `saveMessage:updateSession:` upload trigger (`.m:1376-1393`): trigger message-type membership, and, only when trigger event types are configured, the hashed `name`+`type` lookup. `MPIHasher` is already Swift, so the caller passes one in and no new singleton read crosses the boundary.

Two incidental cleanups: the `NSError *error` the original declared was never assigned, so the `!error` guard is gone; and the caller now reuses `MParticle.getLogger` instead of constructing a second `MPLog`.

**`MPBackendMessageInfo.crashReportBytesToRetainForMessageLength:...`** — the crash-report truncation arithmetic in `logCrash:`. It joins `base64CrashReport:maxBytes:`, which already handles the other half of the same concern. Returns nil when the message already fits. The retained length can still go negative when the rest of the message alone overflows; `truncateMessageDataProperty:toLength:` ignores a negative length so the report is left intact — the original arithmetic behaved the same way, and there is a test pinning it.

**`MPSessionTimingPolicy.cleanUpPlanWithNow:...`** — `cleanUp:`'s due-time check, the `NINETY_DAYS` retention fallback and the next-run time. Both constants stay in `MPIConstants` and are passed in.

## Three planned slices deliberately left in Objective-C

- **`eventWithName:`** filters a set of `MPEvent`. `MPEvent` is Objective-C only and pinned there by the kit ABI (101 files under `Kits/` reference it), so Swift cannot take the set. This one was in the plan before I checked the type.
- **The crash-session pick** and **the `didBecomeActive` launch flag** are a `first(where:)` and a nil check embedded in dictionary assembly. Extracting them adds a boundary for no clarity gain — the same call #924 made for the other trivial `messageInfo` builders.

## Validation

- `Tools/abi-guard.sh check` → exit 0, no baseline or retained-manifest change
- `trunk check` → clean
- iOS build → 0 errors
- Swift suite: **844 passed, 0 failed**, including 13 new mirror tests
- Objective-C contract tests: **165/165** across `MPBackendControllerTests`, `MPPersistenceControllerTests`, `MPUploadBuilderTests`, `MPMessageBuilderTests` — run twice, clean both times

### DIFF
--- a/mParticle-Apple-SDK/MPBackendController.m
+++ b/mParticle-Apple-SDK/MPBackendController.m
@@ -1087,11 +1087,11 @@ static BOOL skipNextUpload = NO;
     MPMessageBuilder *messageBuilder = [[MPMessageBuilder alloc] initWithMessageType:MPMessageTypeCrashReport session:crashSession messageInfo:messageInfo context:self.messageBuilderContext];
     MPMessage *crashMessage = [messageBuilder build];
     
-    NSInteger maxBytes = [MPPersistenceController_PRIVATE maxBytesPerEvent:crashMessage.messageType];
-    if(crashMessage.messageData.length > maxBytes) {
-        NSInteger bytesToTruncate = crashMessage.messageData.length - maxBytes;
-        NSInteger bytesToRetain = plCrashReportBase64.length - bytesToTruncate;
-        [crashMessage truncateMessageDataProperty:kMPPLCrashReport toLength:bytesToRetain];
+    NSNumber *bytesToRetain = [MPBackendMessageInfo crashReportBytesToRetainForMessageLength:crashMessage.messageData.length
+                                                                                    maxBytes:[MPPersistenceController_PRIVATE maxBytesPerEvent:crashMessage.messageType]
+                                                                          base64ReportLength:plCrashReportBase64.length];
+    if (bytesToRetain != nil) {
+        [crashMessage truncateMessageDataProperty:kMPPLCrashReport toLength:bytesToRetain.integerValue];
     }
     [persistence saveMessage:crashMessage];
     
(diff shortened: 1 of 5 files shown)


BrandonStalnaker (dismissed):
Upload still fires on trigger message type, or on hashed name+type only when triggerEventTypes is non-nil. Crash retain length is the same formula (base64Len - (messageLen - max)), including a negative result. Clean-up still uses nextCleanUpTime < now, NINETY_DAYS if max-age is nil, and now + TWENTY_FOUR_HOURS.

mParticle-Apple-SDK/MPBackendController.m
```diff
@@ -1087,11 +1087,11 @@ -  (void)logCrash:(NSString *)message stackTrace:(NSString *)stackTrace plCrashR
...
     
-    NSInteger maxBytes = [MPPersistenceController_PRIVATE maxBytesPerEvent:crashMessage.messageType];
-    if(crashMessage.messageData.length > maxBytes) {
-        NSInteger bytesToTruncate = crashMessage.messageData.length - maxBytes;
-        NSInteger bytesToRetain = plCrashReportBase64.length - bytesToTruncate;
-        [crashMessage truncateMessageDataProperty:kMPPLCrashReport toLength:bytesToRetain];
+    NSNumber *bytesToRetain = [MPBackendMessageInfo crashReportBytesToRetainForMessageLength:crashMessage.messageData.length
+                                                                                    maxBytes:[MPPersistenceController_PRIVATE maxBytesPerEvent:crashMessage.messageType]
+                                                                          base64ReportLength:plCrashReportBase64.length];
+    if (bytesToRetain) {
```
BrandonStalnaker: static analyzer is complaining about this Objective C best practice.

MPBackendController.m:1093:9: warning: Converting a pointer value of type 'NSNumber *' to a primitive boolean value; instead, either compare the pointer to nil or call -boolValue [osx.NumberObjectConversion

thomson-t: Good catch — fixed, and you're right that it's a real gate rather than a style nit.

`if (bytesToRetain)` is now `if (bytesToRetain != nil)`. The value is an `NSNumber *` because the Swift helper returns `NSNumber?` to express "no truncation needed", so the pointer test was genuinely ambiguous to read as well as to the analyzer.

I reproduced the analyzer job locally with the same command and filter chain the workflow uses, and confirmed it both ways:

- before the fix: `MPBackendController.m:1093:9: warning: Converting a pointer value of type 'NSNumber *' ... [osx.NumberObjectConversion]` — 1 surviving warning, so `run-analyzer` fails
- after the fix: 0 surviving warnings, `** ANALYZE SUCCEEDED **`

I also swept the rest of the stack for the same pattern. The only other pointer-as-boolean I added is `if (plan)` in `cleanUp:`, where `plan` is a plain `MPCleanUpPlan *` rather than a boxed number, so `osx.NumberObjectConversion` doesn't apply and `if (obj)` is idiomatic there. Left as is — say the word if you'd rather have it explicit for consistency.

#958 and #959 have been rebased onto this and force-pushed, so the stack is intact.

One process note for my own part: `analyze` isn't one of the five items in `docs/swift-migration/PR-GATE.md`, which is why this reached you instead of being caught locally. I'll run it per-PR from here on; worth considering adding it to the gate.

BrandonStalnaker (approved):

mParticle-Apple-SDK/MPBackendController.m
thomson-t: Thanks. Fixed

PR #958, opened 2026-09-04
refactor: move the user-identity change decision to Swift

thomson-t:
Stacked on #956 (`refactor/swift-backend-decision-slices`).

`setUserIdentity:identityType:timestamp:completionHandler:` decided what to do to the stored identity array inline, across ~70 lines of index lookups, an `NSPredicate`, a block-based `objectTester` and four branches. That decision moves to `MPUserIdentityLogic.planForIdentityType:...`, which answers with an `MPUserIdentityChangePlan`: `unchanged`, `nothingToRemove`, `remove`, `add` or `replace`, plus the index, the entry being displaced and the first-set stamp.

## Why now — this reverses #923's deferral

#923 deliberately left this core in Objective-C, on the grounds that it orchestrates Objective-C objects. What changed is that `MPUserIdentityInstancePRIVATE` and `MPUserIdentityChangePRIVATE` are now fully Swift, so the helper reasons over plain `[[String: Any]]` entries instead of marshalling around SDK objects.

The `.m` still owns every side effect: mutating the array, building the change object, persisting to user defaults, logging, and reporting the exec status.

## Two asymmetries preserved deliberately

Both are in the original, both are now pinned by tests rather than left implicit:

- **The equality check reads the *last* entry of the type; the mutation acts on the *first*.** With well-formed storage there is at most one entry per type and the two agree, but the behaviour on a duplicated type is unchanged.
- **An empty string is *valid* for the equality check yet *removes* in the mutation.** So setting `""` over a stored `""` reports unchanged rather than removing.

`nothingToRemove` is a distinct outcome from `unchanged`: both persist nothing, but the first reports `MPExecStatusSuccess` and the second reports `MPExecStatusFail`, exactly as before.

## Two incidental improvements

- The stored identities are read **once** instead of three times. The original called `userIdentitiesForUserId:` for the predicate lookup, again for the index lookup, and the same user-defaults read returned the same array each time.
- An entry whose type key is not a number is skipped rather than being sent `isEqualToNumber:`, which would previously have raised.

## Validation

Identity is a high-consequence path, so the contract tests are the gate here, not a formality:

- Objective-C: **118/118** across `MPBackendControllerTests`, `MPIdentityTests`, `MPUserIdentityChangeTests` — run twice, clean both times. Includes `testSetIdentityToNil`, `testSetIdentityToNSNull`, `testDoNotSetDuplicateIdentity`, `testDoNotSetDuplicateIdentityCasing`, `testUserIdentityChanged`.
- Broader sweep: **208/208** across `MParticleTests`, `MPPersistenceControllerTests`, `MPKitContainerTests`, with one unrelated failure — `testActiveKitsRegistryThreadSafety`, a self-declared non-deterministic stress test that passes in isolation twice and touches no code in this PR.
- Swift suite: **857 passed, 0 failed**, including 21 new mirror tests.
- `Tools/abi-guard.sh check` → exit 0 · `trunk check` → clean · iOS build → 0 errors.

### DIFF
--- a/mParticle-Apple-SDK-Swift/Sources/Utils/MPUserIdentityLogic.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Utils/MPUserIdentityLogic.swift
@@ -8,8 +8,11 @@ import Foundation
     /// the first non-user identity type (`maxValidTypeExclusive`). Malformed or
     /// out-of-range entries are dropped; order is preserved.
     @objc(validIdentities:typeKey:maxValidTypeExclusive:)
-    public static func validIdentities(_ identities: [[AnyHashable: Any]], typeKey: String,
-                                       maxValidTypeExclusive: Int) -> [[AnyHashable: Any]] {
+    public static func validIdentities(
+        _ identities: [[AnyHashable: Any]],
+        typeKey: String,
+        maxValidTypeExclusive: Int
+    ) -> [[AnyHashable: Any]] {
         identities.filter { entry in
             guard let type = entry[typeKey] as? NSNumber else { return false }
             return type.intValue < maxValidTypeExclusive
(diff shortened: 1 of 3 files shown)


BrandonStalnaker (dismissed):
Thanks for documenting the asymmetry

BrandonStalnaker (approved):

PR #959, opened 2026-09-04
refactor: move the user-attribute mutation decision to Swift

thomson-t:
Stacked on #958 (`refactor/swift-backend-user-identity-diff`). Last PR of this wave.

`setUserAttributeChange:completionHandler:` decided what to do to the stored attributes by **round-tripping through an `NSError`**: it called `+checkAttribute:`, which turns an `MPAttributeValidationResult` into an error code, then branched on that code. The decision now happens on the enum directly.

## What changed

`+checkAttribute:` splits into `+validateAndLogAttributeKey:value:` (validation plus the console messages) and `+errorCodeForValidationResult:`, with `+checkAttribute:` reduced to composing the two. Its signature, error domain and codes are unchanged, and **the logging stays in Objective-C** so `MPILogError`'s level gate keeps deciding whether an offending value is ever interpolated — the trap that cost this project #920.

`MPUserAttributeLogic.mutationForValidationResult:keyExists:` answers `store`, `delete` or `reject`. `valueToLogFor:` performs the number-to-string coercion for the change message. The `.m` keeps the dictionary mutation, the `deletedUserAttributes` set, persistence, logging and the exec statuses.

## Why now — this reverses #922's deferral

#922 left this core in Objective-C because it looked entangled with ivars, SDK objects and gated logging. What makes it tractable is that the decision reduces to two plain values: the validation result, and whether the key is already present. Everything else stays at the boundary.

## Three dead things the restructure removes

All behaviour-preserving, all found while mapping the branches:

- **`if (userAttributeChange.isArray)` had two byte-identical arms.**
- **The early return for `kInvalidDataType`** produced exactly the outcome the final `else` already produced — both collapse into `reject`.
- **`userAttributeValue`** was a local always assigned `userAttributeChange.value`.

## Setter preludes

The two value predicates move to `MPAttributeValidator` as `isAcceptableScalarValue:` and `isAcceptableValueList:` — the first in particular was an easy-to-misread `!((isString && length > 0) || isNumber) && value != nil`.

The shared key check stays in Objective-C. After `[key mutableCopy]` it reduces to a nil test, and its `isKindOfClass:` arm is unreachable for a non-nil key — extracting it would move a nil check behind a call boundary for no gain.

## Validation

- Objective-C: **116/116** across `MPBackendControllerTests` and `MPIdentityTests` — twice, clean both times. Covers the 12 `testCheckAttribute*` tests that pin the split `+checkAttribute:`, plus `testSet*Attribute`, `testRemove*Attribute`, `testIncrement*`, `testUserTag*` and `testUserAttributeChanged*`.
- Full Objective-C target run: only `MPKitContainerTests.testActiveKitsRegistryThreadSafety` fails — a self-declared non-deterministic stress test, and the whole `MPKitContainerTests` suite passes 85/85 on a rerun. Nothing in this PR touches kits.
- Swift suite: **866 passed, 0 failed**, including 17 new mirror tests.
- `Tools/abi-guard.sh check` → exit 0 · `trunk check` → clean · iOS build → 0 errors.

### DIFF
--- a/mParticle-Apple-SDK-Swift/Sources/Utils/MPAttributeValidator.swift
+++ b/mParticle-Apple-SDK-Swift/Sources/Utils/MPAttributeValidator.swift
@@ -68,4 +68,24 @@ import Foundation
 
         return .valid
     }
+
+    /// Whether a scalar attribute value may be passed on for full validation: a non-empty string,
+    /// any number, or nil — a nil value reaches the change path as a removal rather than being
+    /// rejected here.
+    @objc(isAcceptableScalarValue:)
+    public static func isAcceptableScalarValue(_ value: Any?) -> Bool {
+        guard let value else { return true }
+        if let string = value as? NSString { return string.length > 0 }
+        return value is NSNumber
+    }
+
+    /// Whether a list attribute value may be passed on: a non-empty array. Entry types are checked
+    /// later by `validate(key:value:...)`, which reports the offending entry.
+    @objc(isAcceptableValueList:)
+    public static func isAcceptableValueList(_ values: Any?) -> Bool {
+        // `[Any]`, not `NSArray`: the bridged Swift array has `isEmpty`, which SwiftLint's
+        // `empty_count` rule requires and which `NSArray` does not provide.
+        guard let values = values as? [Any] else { return false }
+        return !values.isEmpty
+    }
 }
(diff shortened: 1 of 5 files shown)


BrandonStalnaker (dismissed):
LGTM

BrandonStalnaker (approved):

PR #970, opened 2026-09-08
refactor: cut persistence facade over to Swift

BrandonStalnaker:
## Background
- With all persistence operations implemented in Swift, the existing Objective-C controller can stop owning SQLite while temporarily preserving its selectors and compatibility contracts.

## What Has Changed

- Replaced Objective-C SQL implementations with forwarding calls to MPPersistenceStorePRIVATE.
- Retained the existing Objective-C facade, selectors, runtime behavior, and contract-object marshalling.
- Added compatibility coverage through Objective-C and direct Swift persistence tests.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Objective-C access to persistence and database migration operations.
  * Added support for creating, updating, deleting, and retrieving persisted cookies.

* **Bug Fixes**
  * Upload persistence now consistently respects explicit opt-out status for each upload.
  * Improved handling of consumer information when cookie records are present or absent.
  * Sessionless messages migrated from older versions are now assigned the correct session identifier.

* **Refactor**
  * Centralized database access and migration handling through the persistence store for more consistent behavior.

### DIFF
--- a/mParticle-Apple-SDK/Persistence/MPPersistenceController.m
+++ b/mParticle-Apple-SDK/Persistence/MPPersistenceController.m
@@ -350,17 +350,17 @@
             [rawCookies addObject:[self rawCookie:cookie consumerInfoId:0 mpid:mpid]];
         }
     }
-    consumerInfo.consumerInfoId =
-        [self.store saveRawConsumerInfoForMpid:mpid
-                             uniqueIdentifier:consumerInfo.uniqueIdentifier
-                                      cookies:rawCookies];
-    NSArray<NSDictionary *> *savedCookies = [self.store fetchRawCookiesForUserId:mpid];
+    NSDictionary *savedInfo =
+        [self.store saveRawConsumerInfoWithCookieIdsForMpid:mpid
+                                           uniqueIdentifier:consumerInfo.uniqueIdentifier
+                                                    cookies:rawCookies];
+    consumerInfo.consumerInfoId = [savedInfo[@"consumerInfoId"] longLongValue];
+    NSArray<NSNumber *> *cookieIds = savedInfo[@"cookieIds"];
+    NSUInteger cookieIdIndex = 0;
     for (MPCookie *cookie in consumerInfo.cookies) {
-        for (NSDictionary *saved in savedCookies) {
-            if ([cookie.name isEqualToString:[self nullableString:saved[@"name"]]]) {
-                cookie.cookieId = [saved[@"id"] longLongValue];
-                break;
-            }
+        if (!cookie.expired && cookieIdIndex < cookieIds.count) {
+            cookie.cookieId = cookieIds[cookieIdIndex].longLongValue;
+            cookieIdIndex += 1;
         }
     }
 }
--- a/mParticle-Apple-SDK/Persistence/MPPersistenceController.m
+++ b/mParticle-Apple-SDK/Persistence/MPPersistenceController.m
@@ -5,7 +5,6 @@
 #import "MPConsentState.h"
 #import "MPConsumerInfo.h"
 #import "MPForwardRecord.h"
-#import "MPIntegrationAttributes.h"
 #import "MPPersistenceUploadSettingsCodec.h"
 #import "MPStateMachine.h"
 #import "MPUploadSettings.h"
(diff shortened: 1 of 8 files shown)


thomson-t (dismissed):

mParticle-Apple-SDK/Persistence/MPPersistenceController.m
```diff
@@ -203,1585 +136,296 @@ + (NSInteger)maxBytesPerBatch:(NSString *)messageType {
...
-    
-    NSArray<MPCookie *> *cookies = [self fetchCookiesForUserId:userId];
-    if (cookies.count) {
-        consumerInfo = [[MPConsumerInfo alloc] init];
-        consumerInfo.cookies = cookies;
-    }
-    
-    return consumerInfo;
+- (void)saveUpload:(MPUpload *)upload {
+    self.store.optedOut = [MParticle sharedInstance].stateMachine.optOut;
```
nickolas-dimitrakas: This sets `optedOut` as shared mutable state on the singleton store immediately before the save call, replacing what used to be an atomic inline read of `stateMachine.optOut` at insert time. If `saveUpload:` can ever be invoked from two threads against the same `self.store`, the set-then-use here is racy: one call's flag can be overwritten by the other between this line and the actual insert, silently persisting opted-out data or dropping opted-in data.

Today every call site I found is dispatched onto `[MParticle messageQueue]`, so it may not be reachable in practice — but that's an unenforced assumption. Consider threading `optedOut` through as an explicit parameter on `saveUpload:`/`saveUploads:deleteMessages:` (protocol + Swift-side signature change) instead of a shared property, so correctness doesn't depend on every future caller staying on the same serial queue.

mParticle-Apple-SDK/Persistence/MPPersistenceController.m
```diff
@@ -203,1585 +136,296 @@ + (NSInteger)maxBytesPerBatch:(NSString *)messageType {
...
-        if (sqlite3_step(preparedStatement) == SQLITE_ROW) {
-            integrationAttributes = [[MPIntegrationAttributes alloc] initWithIntegrationId:integrationId
-                                                                            attributesData:dataValue(preparedStatement, 0)];
-        }
-    }
-    
-    sqlite3_finalize(preparedStatement);
-    return [integrationAttributes attributes];
+- (BOOL)saveUploads:(NSArray<MPUpload *> *)uploads deleteMessages:(NSArray<MPMessage *> *)messages {
+    self.store.optedOut = [MParticle sharedInstance].stateMachine.optOut;
```
nickolas-dimitrakas: Same shared-mutable-state concern as `saveUpload:` above — `self.store.optedOut` is set here and read again inside `saveUploads:deleteMessages:`, with no atomicity between the two. Same suggestion: pass `optedOut` as an explicit parameter rather than relying on a property set right before the call.

mParticle-Apple-SDK/Persistence/MPPersistenceController.m
```diff
@@ -203,1585 +136,296 @@ + (NSInteger)maxBytesPerBatch:(NSString *)messageType {
...
-        sqlite3_bind_int(preparedStatement, 3, (int)[MPPersistenceSchemaPRIVATE maxBreadcrumbs]);
-        
-        if (sqlite3_step(preparedStatement) != SQLITE_DONE) {
-            MPILogError(@"Error while pruning breadcrumbs: %s", sqlite3_errmsg(mParticleDB));
-        }
-        
-        sqlite3_clear_bindings(preparedStatement);
-    } else {
-        MPILogError(@"could not prepare statemnt: %s\n", sqlite3_errmsg(mParticleDB));
+- (MPConsumerInfo *)fetchConsumerInfoForUserId:(NSNumber *)userId {
```
nickolas-dimitrakas, line 335: This now populates `consumerInfo.consumerInfoId` and `.uniqueIdentifier` from `fetchRawConsumerInfoForUserId:` (lines 342-343). The previous ObjC implementation of this method never read the `consumer_info` row at all — it only ever built `MPConsumerInfo` from `cookies` and left `consumerInfoId`/`uniqueIdentifier` at their defaults. That's a real, user-visible behavior change (it flows into the public, read-only `MParticle.uniqueIdentifier` property), which contradicts this PR's description of "retained... runtime behavior."

This may well be an intentional fix riding along with the port — if so, worth calling out explicitly in the PR description with before/after behavior, rather than shipping silently as part of a "no behavior change" migration PR.

nickolas-dimitrakas (changes requested):
Two findings from a full read of the diff (details inline below):

1. `optedOut` moved from an atomic point-of-read (`stateMachine.optOut` read inline at insert time) to a shared mutable property set on the singleton `MPPersistenceStorePRIVATE` immediately before the save call. As far as I can tell every current caller of `saveUpload:`/`saveUploads:deleteMessages:` happens to run on `[MParticle messageQueue]`, so I don't have a concrete repro of two calls interleaving today — but nothing enforces that invariant, so a future caller off that queue (or a queue-hopping refactor) would silently corrupt opt-out enforcement. Worth hardening now while the call sites are still small in number.
2. `fetchConsumerInfoForUserId:` now populates `consumerInfoId`/`uniqueIdentifier` from `fetchRawConsumerInfoForUserId:`, which the old ObjC implementation never did (it only ever set `.cookies`). That's a real behavior change, which cuts against the PR description's "retained... runtime behavior" claim. It flows into the public `MParticle.uniqueIdentifier` getter. Could well be an intentional bug fix — please just confirm explicitly so it's not shipped as an unreviewed side effect.

Requesting changes so these get an explicit yes/no rather than silently merging.

mParticle-Apple-SDK/Persistence/MPPersistenceController.m
BrandonStalnaker: Good catch—this was an unintended behavior change. I restored the legacy behavior: the method only returns cookie-backed consumer info and leaves consumerInfoId and uniqueIdentifier unset. Regression tests were added to cover this.

BrandonStalnaker:
Thanks Nick both concerns were valid and are addressed in 7300fe1d.

Opt-out state is now passed explicitly to saveUpload and saveUploads, so enforcement no longer depends on mutable state stored between a setter and save call.
fetchConsumerInfoForUserId: now preserves the previous Objective-C behavior: it returns consumer information only when cookies exist and does not hydrate consumerInfoId or uniqueIdentifier from the database row. This also avoids double-encoding persisted identifiers.
I added regressions for per-save opt-out isolation, cookie-only restoration, and ignoring consumer rows without cookies. The changes were propagated through the remaining stacked branches. Targeted tests, lint, and static analysis pass.

nickolas-dimitrakas (approved):

PR #972, opened 2026-09-08
refactor: inject session and message persistence

BrandonStalnaker:
## Background
- Backend session and message operations still reach persistence through the Objective-C singleton facade despite the injected Swift persistence dependency.

## What Has Changed

- Migrated session begin, end, recovery, archive, and update paths to injected persistence.
- Migrated the central message and breadcrumb save paths.
- Removed singleton persistence access from these backend lifecycle operations.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved internal persistence handling for sessions, messages, uploads, breadcrumbs, and cleanup operations.
  * No user-visible behavior changes.

### DIFF
--- a/mParticle-Apple-SDK/MPBackendController.m
+++ b/mParticle-Apple-SDK/MPBackendController.m
@@ -217,7 +217,7 @@ const NSTimeInterval kMPRemainingBackgroundTimeMinimumThreshold = 10.0;
 }
 
 - (void)confirmEndSessionMessage:(MPSession *)session {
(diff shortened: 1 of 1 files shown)


nickolas-dimitrakas (dismissed):
Approving. Noting one issue of its own: retyping `persistence` to `id<MPBackendPersistence>` newly triggers a `-Wunused-result` warning on `archiveSession:` (run-analyzer fails here as a result). That warning is cleaned up with a `(void)` cast in #975, so not blocking, but flagging so it's not lost.

nickolas-dimitrakas (approved):

PR #974, opened 2026-09-08
refactor: migrate peripheral persistence callers

BrandonStalnaker:
## Background
- Production callers outside the backend still depend on MPPersistenceController, preventing removal of the Objective-C facade.

## What Has Changed
- Migrated upload builder, identity, kits, notifications, state machine, network communication, and SDK-level callers.
- Added a narrow Objective-C adapter for contract types that cannot be constructed directly by the Swift target.
- Moved MPID, consent, and schema utility access out of the persistence facade.
- Updated affected tests and mocks.

## Checklist

### DIFF
--- a/mParticle-Apple-SDK/Utils/MPUserDefaultsConnector.m
+++ b/mParticle-Apple-SDK/Utils/MPUserDefaultsConnector.m
@@ -71,7 +71,8 @@
 }
 
 + (MPConsentState *)effectiveConsentStateForMpid:(NSNumber *)mpid {
-    return self.deviceConsentState ?: (mpid ? [self consentStateForMpid:mpid] : nil);
+    return self.deviceConsentState
+        ?: (mpid != nil ? [self consentStateForMpid:mpid] : nil);
 }
 
 + (NSInteger)maxBytesPerEvent:(NSString *)messageType {
--- a/mParticle-Apple-SDK/Utils/MPUserDefaultsConnector.m
+++ b/mParticle-Apple-SDK/Utils/MPUserDefaultsConnector.m
@@ -1,6 +1,10 @@
 #import "MPUserDefaultsConnector.h"
 #import "mParticle.h"
 #import "../Kits/MPKitContainer+MParticlePrivate.h"
+#import "MPPersistenceUtilities.h"
+#import "MPIConstants.h"
+#import "MPConsentSerialization.h"
+#import "MPConsentState.h"
 
 @import mParticle_Apple_SDK_Swift;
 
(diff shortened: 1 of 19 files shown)


mParticle-Apple-SDK/Utils/MPUserDefaultsConnector.m
```diff
@@ -16,6 +20,70 @@ @interface MPUserDefaultsConnector()<MPUserDefaultsConnectorProtocol>
...
+        if (!string) {
+            return;
+        }
+        [userDefaults setMPObject:string forKey:kMPConsentDeviceStateKey userId:@0];
+    }
+    [userDefaults synchronize];
+}
+
++ (MPConsentState *)effectiveConsentStateForMpid:(NSNumber *)mpid {
+    return self.deviceConsentState ?: (mpid ? [self consentStateForMpid:mpid] : nil);
```
nickolas-dimitrakas: New clang analyzer warning here (`osx.NumberObjectConversion`): `mpid` is an `NSNumber *`, so `mpid ? ... : nil` reads as truthiness-of-value to the checker even though the intent is a plain nil-check. This is what fails `run-analyzer` on this PR. Same fix already lands in #975 — just needs to happen here at the PR that introduces the method:

```suggestion
    return self.deviceConsentState ?: (mpid != nil ? [self consentStateForMpid:mpid] : nil);
```

nickolas-dimitrakas (changes requested):
Root-caused both CI failures on this PR from the raw job logs rather than just reading the red X:

1. `MPIdentityTests.testMPIdZeroToMPId` (UnitTests/ObjCTests/MPIdentityTests.m, ~line 809) fails deterministically (not the known kitsRegistry flake) because this PR reroutes `MPIdentityApi.m:228` from `[MParticle sharedInstance].persistenceController` to `.persistenceAdapter`, but that pre-existing test only stubs `persistenceController` on its `MParticle` class mock: `[[[mockInstance stub] andReturn:mockPersistenceController] persistenceController];`. `persistenceAdapter` is left unstubbed, so the OCMock class mock returns nil for it, `moveContentFromMpidZeroToMpid:` is sent to nil, and `mockPersistenceController`'s `expect` never fires. This is a test-instrumentation gap this PR's own caller migration leaves behind (not a production bug — production's real `persistenceAdapter` correctly forwards to the real controller). I can't leave an inline comment on that line since this PR's diff doesn't touch that file, but the fix is a one-line addition right after the existing stub: `[[[mockInstance stub] andReturn:mockPersistenceController] persistenceAdapter];` (matches the pattern already used for this elsewhere in the stack, e.g. `((MParticle *)mockInstance).persistenceAdapter = mockPersistenceController;`).
2. `run-analyzer` fails on a new warning in `MPUserDefaultsConnector.m` (`osx.NumberObjectConversion`) from treating an `NSNumber *` as a boolean. See inline comment/suggestion — same fix already applied downstream in #975, just needs to happen here at the source.

Requesting changes to fix both at the PR that introduces them rather than carrying them further down the stack.

nickolas-dimitrakas (approved):

PR #975, opened 2026-09-08
refactor: remove Objective-C persistence facade

BrandonStalnaker:
## Background
- After migrating all production callers, the Objective-C persistence facade is obsolete. Startup can now create and own the Swift store directly.
- Removing the facade prevents persistence ownership from remaining split between Objective-C and Swift.

## What Has Changed
- Moved persistence initialization, migration, and ownership into SDK startup.
- Applied credentials and network options before persistence migration.
- Injected the Swift store directly into the backend.
- Made persistence accessors initialize lazily while preserving injected test implementations.
- Removed `MPPersistenceController.m`, its header, project references, and obsolete facade tests.
- Retained a narrow Objective-C contract adapter where Swift cannot construct Objective-C runtime types.
- Updated the ABI baseline and remaining tests for the new persistence architecture.

## Objective-C Contract Audit
- Before regenerating the ABI baseline, its diff was limited to the audited internal persistence surface:
  - `MPPersistenceController_PRIVATE`
  - `MPPersistenceControllerProtocol`
  - Their database, session, message, upload, and peripheral persistence declarations
  - Their internal replacements: `MPPersistenceAdapter`, `MPPersistenceAdapting`, and `MPPersistenceUtilities`
- Customer API audit: the removed controller and protocol were internal implementation types and were not exposed through the documented `MParticle` customer API.
- Kit audit: no kit-facing protocol, selector, or runtime contract depends on the removed types.
- Wrapper-SDK audit: no wrapper SDK references or imports the removed controller surface.
- Runtime and persistence audit: the removed types are not securely archived, restored by runtime name, or encoded in persisted data. SQLite filenames, schemas, migration behavior, and archived upload settings remain unchanged.
- The baseline was regenerated from the final tree, and `bash Tools/abi-guard.sh check` passes with no public ABI diff.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Reworked the SDK’s persistence layer around a persistence adapter and store.
  * Updated database-backed handling for sessions, uploads, cookies, consumer information, integration attributes, and forward records.
  * Persistence initialization is now handled automatically during SDK startup.
  * Consent state and user identifier utilities have updated interfaces with clearer nullability guarantees.
  * Replaced the previous persistence controller interface with the new adapter-based interface.


mParticle-Apple-SDK/mParticle.m
```diff
@@ -165,6 +195,42 @@ - (instancetype)init {
     return self;
 }
 
+- (void)initializePersistence {
+    if (_persistenceStore) {
```
nickolas-dimitrakas: This guard is why both `MParticleConfigurationTests.test_reset_clearsState_andFlushesKits` and `MParticleKitBatchTests.test_logKitBatch_executesKitHandler_andPersistsForwardRecords` fail deterministically (not flake) on this PR.

`persistenceAdapter`/`persistenceStore` are declared as mutable `@property (nonatomic, strong)` specifically so tests can inject fakes directly (`MParticleTestBase.swift:70`: `mparticle.persistenceAdapter = persistenceController`). But this method guards solely on `_persistenceStore`. When a test sets `_persistenceAdapter` to a fake without also having `_persistenceStore` populated, the very next access to `self.persistenceAdapter` (or `.persistenceStore`) calls this method again, sees `_persistenceStore` is still nil, and reruns the whole body — silently overwriting the injected `_persistenceAdapter` with a brand-new real one. The test's mock is discarded and production code ends up talking to a real adapter instead, which is exactly why `resetDatabaseCalled`/`saveCalled` never flip to true on the fakes.

```suggestion
    if (_persistenceStore || _persistenceAdapter) {
        return;
    }
```
This isn't only a test-injection nicety — it means *any* code path that sets `persistenceAdapter` before persistence has otherwise been touched loses that assignment the next time the property is read. Worth the one-line guard fix.

Tools/abi-baseline.txt
```diff
@@ -248,7 +248,8 @@ IFACE|@interface MPModifyApiResult : MPIdentityApiResult
 IFACE|@interface MPNetworkCommunication_PRIVATE : NSObject
 IFACE|@interface MPNetworkOptions : NSObject
 IFACE|@interface MPNotificationController_PRIVATE : NSObject <MPNotificationControllerProtocol>
-IFACE|@interface MPPersistenceController_PRIVATE : NSObject<MPPersistenceControllerProtocol>
+IFACE|@interface MPPersistenceAdapter : NSObject <MPPersistenceAdapting>
```
nickolas-dimitrakas, line 251: Per `docs/swift-migration/PR-GATE.md` item 1, a baseline update must name the audited symbols and summarize compatibility evidence in the PR body — that step isn't in this PR's description. The content itself checks out (everything removed traces to the deleted `MPPersistenceController_PRIVATE`/`MPPersistenceControllerProtocol` internal surface, nothing customer/kit/wrapper-SDK-facing), but please add the audit note to the PR description so this baseline change is on the record per the policy, rather than an unexplained diff.

nickolas-dimitrakas (changes requested):
Root-caused both failing tests on this PR (`MParticleConfigurationTests.test_reset_clearsState_andFlushesKits` and `MParticleKitBatchTests.test_logKitBatch_executesKitHandler_andPersistsForwardRecords`, both deterministic on iOS and tvOS, no flake signature) to a single bug in the new `initializePersistence` lazy-init guard — see inline comment on `mParticle.m`. Also flagging the ABI baseline update, which is safe in content but is missing the audit write-up the migration gate requires.

Requesting changes for the `initializePersistence` fix (it's a real bug, not just a test gap) and to add the audit note for the baseline change.

Tools/abi-baseline.txt
BrandonStalnaker: updated the PR message to follow PR-GATE.md

nickolas-dimitrakas (approved):

PR #978, opened 2026-09-08
refactor: migrate MPCookie to Swift

thomson-t:
Deletes the `MPCookie` wrapper from `mParticle-Apple-SDK/Data Model/MPConsumerInfo.m` (143 LOC) and gives the existing `MPCookiePRIVATE` the `@objc(MPCookie)` runtime name, so `Include/MPPersistenceController.h`'s `@class` forward declaration and its `-deleteCookie:` / `-fetchCookiesForUserId:` signatures stay byte-identical.

`MPCookiePRIVATE` already owned every accessor, `initWithName:configuration:`, `expired`, `dictionaryRepresentation` and the name comparison. The wrapper's remaining content was `isEqual:`, `hash` and `NSSecureCoding`, which move across as-is.

## The `kMPCK*` wire keys stay Objective-C

Swift cannot emit C globals, and Objective-C callers still reference them by name — `MParticleTests.testPersistenceAdapterAssignsDistinctIdsToSameNameCookies`, added in #975, builds cookie configurations from them. So the three `extern`s stay in `MPConsumerInfo.h`. This module cannot read them, so `MPCookie.swift` keeps private literal copies, the same duplication `MPDevice` already uses; a comment on each side points at the other.

## Scope: `MPConsumerInfo` is not part of this

The two classes share a header and a `.m`, but only `MPCookie` is a clean move. `MPConsumerInfo` still reads `MPPersistenceAdapter` and `MPUserDefaultsConnector.userDefaults` in four of its ten members, and it is constructed at 14 sites plus `initWithCoder:` — which has no dependency to hand it. This repo's injection pattern is constructor injection (`MPResponseConfig.swift:142`, `MPUploadResponseHandler.swift:40`) and it does not reach that class without rewriting how five Objective-C tests obtain their dependencies. That is a redesign of dependency acquisition, not a behaviour-preserving migration, so `MPConsumerInfo` stays in Objective-C and keeps the file and header alive.

Its pure logic is already Swift: escaping, expiry, `dictionaryRepresentation`, the name comparison.

## The cookie coder is ported verbatim, defect included

`content`, `domain` and `expiration` are encoded as strings but decoded with `NSDictionary` as the expected class, so a restored cookie carries only its name:

```objc
[coder encodeObject:self.content forKey:@"content"];                         // NSString in
...
value = [coder decodeObjectOfClass:[NSDictionary class] forKey:@"content"];  // always nil out
```

**No production path archives a cookie** — cookies are written as raw sqlite columns, and the only production archives in the SDK are `MPUploadSettings` and a configuration dictionary. Cookies are archivable in principle, though: `MPConsumerInfo` conforms to `NSSecureCoding` and encodes its `cookies` array, and `MPConsumerInfoTests.testInstance` and `testConsumerInfoEncoding` do exercise that path. Neither asserts the cookie fields, and `isEqual(toCookie:)` compares names only, which is why the loss goes unnoticed there and in `testCookie`'s `XCTAssertEqualObjects(cookie, deserializedCookie)`.

Repairing it would change what a decoded cookie contains, which is a behaviour change rather than a migration, so it is left alone and pinned by a test that states the loss explicitly. Fixed separately in #979. (Contrast with #977, where the equivalent defect made the decode path *impossible* rather than lossy, which forced the fix into that PR.)

## New Swift coverage

In `MPDataModelTests`: runtime-name identity, name-based equality including the foreign-type and nil rejections `testCookie` asserts from Objective-C, the nil `dictionaryRepresentation`, and the archive round trip above.

`UnitTests/Mocks/MPPersistenceControllerMock.swift` reaches the type by its Swift name, `MPCookiePRIVATE`, now that it is no longer an Objective-C class behind the bridging header.

## Compatibility

- **No ABI change** — `bash Tools/abi-guard.sh check` exits 0, baseline untouched.
- **No manifest change.**
- **No pbxproj change** — the file the wrapper lived in still exists.

## Verification

Rebased onto `workstation/swift-migration` after #975 (`remove Objective-C persistence facade`) and #977 landed, and re-run end to end:

- `bash Tools/abi-guard.sh check` → exit 0
- `trunk check` → clean (re-checked after committing)
- iOS build → clean, no new warnings
- `run-analyzer`'s exact filter chain → **0 surviving warnings**
- Objective-C suite → green; Swift suite → green


mParticle-Apple-SDK-Swift/Sources/DataModel/MPCookie.swift
```diff
@@ -60,19 +65,73 @@ import Foundation
...
+
+        if let content { coder.encode(content, forKey: CodingKeys.content) }
+        if let domain { coder.encode(domain, forKey: CodingKeys.domain) }
+        if let expiration { coder.encode(expiration, forKey: CodingKeys.expiration) }
+    }
+
+    /// Ported verbatim from the deleted Objective-C wrapper, including a defect: `content`,
+    /// `domain` and `expiration` are encoded as strings but decoded with `NSDictionary` as the
+    /// expected class, so they never survive a round trip and a restored cookie carries only its
+    /// name. Nothing in the SDK archives a cookie — persistence writes raw sqlite columns — and
```
BrandonStalnaker:
```suggestion
    /// name. Nothing in the production persistence path. — persistence writes raw sqlite columns — and
```
Nit: “Nothing in the SDK archives a cookie” is a bit strong. Persistence is sqlite, yes, but MPConsumerInfo still implements NSSecureCoding and does [coder encodeObject:_cookies forKey:@"cookies"]. testInstance / testConsumerInfoEncoding already archive that object. Worth tightening the comment so it says production persistence does not use this path, not that nothing archives a cookie.

BrandonStalnaker (dismissed):
Looks good to me

mParticle-Apple-SDK-Swift/Sources/DataModel/MPCookie.swift
thomson-t: Validated and you're right — changed.

`MPConsumerInfo` does `[coder encodeObject:_cookies forKey:@"cookies"]`, and both `MPConsumerInfoTests.testInstance` (`NSKeyedArchiver archivedDataWithRootObject:`) and `testConsumerInfoEncoding` (the secure helper) archive that object, so cookies genuinely do go through the coder. "Nothing in the SDK archives a cookie" was wrong.

What I could verify is the narrower claim: no *production* path archives one. Grepping every `NSKeyedArchiver`/`NSKeyedUnarchiver` site in `mParticle-Apple-SDK` and `mParticle-Apple-SDK-Swift` gives only `MPUploadSettings` and a configuration dictionary — cookies are written as raw sqlite columns. So the comment now says that, and adds why the tests don't catch it either: neither asserts the cookie fields, and `isEqual(toCookie:)` compares names only.

Rewritten in both the source comment and the test comment, and in the commit message.

Two other things this round, since the branch also needed restacking onto #975:

- `#975` added `MParticleTests.testPersistenceAdapterAssignsDistinctIdsToSameNameCookies`, which uses `kMPCKContent`/`kMPCKDomain`/`kMPCKExpiration`. My original audit found zero references and dropped them; that was true at `c2d7ada2` and is not any more. **The three constants are restored** as C globals in `MPConsumerInfo.h`/`.m` — Swift can't emit C globals, so `MPCookie.swift` keeps private literal copies, the same duplication `MPDevice` already uses, with a comment on each side pointing at the other.
- `UnitTests/Mocks/MPPersistenceControllerMock.swift` (also new in #975) referenced `MPCookie` from Swift; it now uses the Swift name `MPCookiePRIVATE`.

Both suites, `abi-guard`, `trunk check` and the `run-analyzer` filter are green on the rebased branch.

BrandonStalnaker (approved):

PR #982, opened 2026-09-09
fix: merge main into workstation/swift-migration and port pending concurrency fixes

nickolas-dimitrakas:
Merges `main`'s tip (23 commits since the last sync in #915) into `workstation/swift-migration`, following the same real-merge approach as #915: a `git merge`, not a hand-authored port, even though the branch's squash-merge-only ruleset means the result still lands as one squashed commit and this same set of conflicts will resurface on the next sync.

## What's in main's 23 commits

Releases v9.4.1-v9.4.3, a GoDaddy TLS Root CA - R1 certificate-pinning addition (#879), CI stability fixes (#867, #870), CODEOWNERS/docs housekeeping, dependency bumps, and - the two commits that mattered most here - **#878** and **#916**, the kitsRegistry concurrency fixes that #915 explicitly flagged as "still open... once fixed there, it reaches this branch on the next sync."

## Porting #878/#916 into the Swift-migrated container

`MPKitContainer.m`/`.h` don't exist on this branch (migrated to `MPKitContainer.swift` + `MPKitContainerExecutionAdapter.m` back in #895, well before #878/#916 existed on main), so the merge treated them as modify/delete conflicts. I kept the deletion and instead ported the *substance* of both fixes into `MPKitContainerExecutionAdapter.m`, since that file still carries the exact pre-fix code shape #878/#916 were written against:

- **`bracketsSemaphore`**, separate from `kitsSemaphore`, guarding `bracketForKit:`/`updateBracketsWithConfiguration:` (closes the same unguarded-dictionary-mutation race #878 fixed).
- **`freeKitRegister:integrationId:`** + **`teardownDetachedWrapperInstance:forIntegrationId:`**, splitting `wrapperInstance`'s detach (inline, under `kitsSemaphore`, since that's what synchronizes it against `activeKitsRegistryWhenLocked`'s reads) from its teardown (stop/disk I/O/notification, deferred off the lock).
- **`kitTeardownGroup`**, exposed as `enterKitTeardownGroup`/`leaveKitTeardownGroup`/`notifyWhenKitTeardownComplete:block:` - added to `MPKitContainer.swift` (alongside the registry it's scoped next to) rather than the ObjC adapter, since `mParticle.m`'s `resetForSwitchingWorkspaces:`/`reset:` already auto-merged calls to `[MPKitContainer_PRIVATE notifyWhenKitTeardownComplete:...]` from main - without this the merge wouldn't have compiled at all.

**A regression I introduced and caught before pushing:** porting the `flushSerializedKits` fix means `wrapperInstance` is now nil'd out *before* `removeAllSideloadedKits` runs (same ordering main hit in #916's third commit). Workstation's `removeAllSideloadedKits` still identified sideloaded kits via `[wrapperInstance respondsToSelector:@selector(sideloadedKitCode)]`, which stopped matching once `wrapperInstance` was already nil. `testSwitchWorkspaceSideloadedKits` caught it immediately (registry count 2 instead of 1 after a workspace switch). Fixed the same way main did: identify by `code >= sideloadedKitCodeStartValue` instead, which survives the detach.

**Deliberately not ported:** the `kitsRegistry`-level lock fix (#916's first commit, "kitsSemaphore was per-instance, kitsRegistry is class-level"). That specific bug can't recur here - the registry itself already lives behind a `static NSLock` in `MPKitContainer.swift`, and every access goes through it. But `kitsSemaphore` in `MPKitContainerExecutionAdapter` *is* still instance-scoped, and a fresh adapter instance is created on every `reset`/workspace switch (`mParticle.m:558`) while all instances share the same static registry - so the *same shape* of bug (an instance-scoped lock guarding what's actually shared state) plausibly still applies to `wrapperInstance`/`brackets` synchronization across adjacent adapter instances during a switch. I didn't find a concrete repro and didn't want to under-test a deeper lock refactor into a routine sync PR - flagging it as a follow-up rather than fixing it here, the same way #915 flagged its own weak-wait finding.

**That weak-wait finding, incidentally, is what this sync fixes.** #915 said Cursor flagged `testSwitchWorkspaceOptions` and `testSwitchWorkspaceSideloadedKits` waiting on signals (`options != nil`, bare registry count) that fire before the SDK actually finishes starting, and that the real fix was on #878/#916. It was: main's version of both tests waits on `.initialized` in addition to the original signal. Took main's condition, kept this branch's required `id<MPExtensionKitProtocol>` cast (the Swift-bridged `registeredKits` returns a bare `NSSet`, not the generic-typed one the deleted ObjC header declared).

## Other conflicts

- **Kit/version files** (`VERSION`, `CHANGELOG.md`, all podspecs, `Framework/Info.plist`, `MPIConstants.m`): mechanical - took main's `9.4.3`.
- **`MPKitRokt.m`**: main's side was the pre-Swift-migration ObjC implementation, superseded by this branch's `@import mParticle_Rokt_Internal` rewrite (#940); main only bumped a version constant inside it. Took this branch's version entirely.
- **`MPConnector.m`/`.h`**: deleted here (migrated to Swift); main's only change since the merge-base was the GoDaddy R1 cert addition, already present in `MPPinnedCertificates.swift`. Kept the deletion, but ported main's regression test (`testDefaultPinnedCertificatesContainGoDaddyTLSRootR1`, verifying the pinned cert's SHA-256 fingerprint) to `MPConnectorTests.swift`, since nothing on this branch was guarding against that cert accidentally changing.
- **`MPConnectorTests.m`/`MPRoktTests.m`**: main added tests and timeout-constant renames against ObjC APIs this branch's Swift migrations already removed (`MPConnector`'s `defaultPinnedCertificates`/`urlSession`/delegate methods; `MPRokt`'s `getRoktPlacementAttributesMapping`/`confirmUser:user:completion:`, now behind `mParticle_Rokt_Internal`). Where this branch already had an equivalent fix under different names (`kMPRoktForwardTimeout`/`drainPendingAsyncWork` vs main's `kMPRoktAsyncTimeout`/`mp_drainSDKWork` - both raise the same 0.2s timeout budget and drain queued work before tearing down mocks), kept this branch's version. Where main added whole tests for now-removed APIs, dropped them.

## Validation

- `xcodebuild build`, both podspec targets: clean.
- ObjC suite (`mParticle-Apple-SDK` scheme): **970/970 passing**, including the ported `testActiveKitsRegistryThreadSafety`/`testBracketForKitThreadSafety` stress tests and all five workspace-switch tests.
- Swift suite (`mParticle-Apple-SDK-Swift` scheme): 893/894 passing. The one failure, `MPPersistenceStoreTests.testMigratesVersion30RowsAndPurgesExpiredRecords`, reproduces identically on `workstation/swift-migration`'s tip in a clean worktree with none of this PR's changes - confirmed pre-existing, unrelated to this merge, not touched here.
- `trunk check`: clean, no new issues.


nickolas-dimitrakas:
Superseded by #983 — renaming this branch to follow semantic-commit naming closed this PR rather than retargeting it, so re-opened as a fresh PR from the renamed branch (`fix/merge-main-into-workstation-swift-migration`).

PR #988, opened 2026-09-11
refactor: migrate MPConsumerInfo to Swift

thomson-t:
refactor: migrate MPConsumerInfo to Swift

`MPConsumerInfo` was already a forwarding wrapper over `MPConsumerInfoPRIVATE`, which held only
`consumerInfoId` and `uniqueIdentifier`. The remaining logic moves to
`Sources/DataModel/MPConsumerInfo.swift` and the Objective-C `.h`/`.m` are deleted (-249 lines of
in-scope Objective-C). The Swift class takes `@objc(MPConsumerInfo)`, so the runtime name, every
selector and the archived class name are unchanged and `Tools/abi-baseline.txt` does not move.

Wave 4 declined this type; reversing that deferral is justified by what has landed since:

- The three dependencies cited as blockers are Swift now. `MPCookie` became `MPCookiePRIVATE`
  (#978), the persistence store is `MPPersistenceStorePRIVATE` (#965-#975), and `MPUserDefaults`
  has been Swift throughout. The Swift class reaches only Swift types.
- The "14 construction sites" count included `#import` lines and tests. There are **two**
  production construction sites: `MPStateMachine.m` and `MPPersistenceAdapter.m`.

`-updateWithConfiguration:` gains an `existingCookies:` parameter rather than a persistence seam.
The method merged the response onto already-persisted cookies by reaching
`[MParticle sharedInstance].persistenceAdapter`, which the Swift module cannot import. Its one
production caller, `+[MPNetworkCommunication_PRIVATE parseConfiguration:]`, already holds the
adapter on the following line, so it does the fetch and passes the cookies in. That is the
value-context pattern `MPURLRequestContext` and `MPMessageBuilderContext` already use, and it
leaves the data model with no persistence reference at all. The nine test call sites pass `nil`;
none of them exercised the merge, and being explicit removes a dependency on whatever the shared
test database happened to hold.

`kMPCKContent` / `kMPCKDomain` / `kMPCKExpiration` move to `MPIConstants.{h,m}`. Swift cannot emit
C globals and `MParticleTests` still builds a cookie configuration from them.

Two behaviours worth calling out:

- The decoder's allowed-class set now names `MPCookiePRIVATE` as well as `NSArray`. The wrapper
  passed `decodeObjectOfClass:[NSArray<MPCookie *> class]`, whose generic erases, so the set held
  `NSArray` alone. A scratch harness showed what that costs: `NSCoder` only enforces the set when
  `requiresSecureCoding` is true, so every path the SDK and its tests use is unaffected, but under
  `unarchivedObject(ofClass:)` the nested cookies set `NSCoder.error` and the *whole* unarchive
  fails rather than just dropping the cookies. Naming both classes leaves every current caller's
  result identical, so this is not a behaviour change needing its own entry - it only repairs a
  path that has no caller. `-[MPBaseTestCase attemptSecureEncodingwithClass:Object:]` is a
  misnomer: it uses the legacy non-secure archiver and asserts only `uniqueIdentifier`, which is
  why nothing caught this.
- `updateWithConfiguration:` and `configureCookiesWithDictionary:` now type-check their argument
  before use. The wrapper guarded only against nil and `NSNull`, so a non-dictionary would have
  crashed on `-allKeys`.

Also removes a dead `MPConsumerInfo(Tests)` category that declared `-generateMpId`, a method with
no implementation anywhere and no callers.

Verified: `abi-guard.sh check` exits 0 with the baseline untouched, `trunk check` reports no new
issues, the `run-analyzer` filter chain yields 0 surviving `: warning:` lines, and both suites are
green - 913 Objective-C tests including the five unchanged `MPConsumerInfoTests` assertions, plus
nine new Swift mirror tests. tvOS was not built locally (the platform is not installed); CI covers
it.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

- **New Features**
  - Improved consumer information and cookie handling, including preservation and updating of existing cookies.
  - Added secure encoding and restoration for consumer information data.
  - Added support for migrating legacy configuration values and maintaining device application details.
  - Added access to previously initialized user preferences and standardized cookie configuration constants.

- **Bug Fixes**
  - Improved cookie configuration behavior for empty, null, and partially configured values.

- **Tests**
  - Expanded coverage for cookie updates, persistence, decoding, identifier handling, and configuration edge cases.


mParticle-Apple-SDK-Swift/Sources/DataModel/MPConsumerInfo.swift
```diff
@@ -0,0 +1,211 @@
...
+
+        let key = RemoteConfig.kMPRemoteConfigCookiesKey
+        guard let localCookies = userDefaults.mpObject(
+            forKey: key,
+            userId: MPUserDefaults.storedMpId()
+        ) as? [AnyHashable: Any] else {
+            return nil
+        }
+
+        userDefaults.removeMPObject(forKey: key)
```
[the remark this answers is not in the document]
thomson-t: Not applicable — the two user IDs are the same object by construction, so the read and the removal cannot diverge.

`MPUserDefaults.removeMPObject(forKey:)` resolves through `connector.mpId()`, and `MPUserDefaultsConnector` implements that as `return [MPPersistenceUtilities mpId];` (`MPUserDefaultsConnector.m:265-267`), which is `return MPUserDefaults.storedMpId;` (`MPUserDefaultsConnector.m:36-38`). So `connector.mpId()` *is* `MPUserDefaults.storedMpId()` — the same call this code passes to `mpObject(forKey:userId:)`.

This is also the deleted wrapper's exact shape: it read with `[MPPersistenceUtilities mpId]` and removed with the one-argument `removeMPObjectForKey:`. Behaviour-preserving migration, so the asymmetry in spelling is deliberately carried over rather than introduced here.

Separately, on the "drain legacy cookies when `ck` is absent" point in the summary: the deleted `-configureCookiesWithDictionary:` also returned on `MPIsNull(cookiesDictionary)` *before* reaching `-localCookiesDictionary` (see the pre-change file at `origin/workstation/swift-migration`). Making an absent `ck` drain the legacy table would be a behaviour change, so it is out of scope here — worth raising as its own issue if the migration of those cookies matters.

mParticle-Apple-SDK-Swift/Sources/DataModel/MPConsumerInfo.swift
```diff
@@ -0,0 +1,211 @@
...
+            guard let newValue, !MPSwiftIsNull(newValue) else { return }
+            storedUniqueIdentifier = newValue.percentEscape()
+        }
+    }
+
+    /// The `g` query parameter carried inside the `uid` cookie's content, or a fresh UUID when
+    /// there is no such cookie. Computed once and then persisted, so the stamp is stable for the
+    /// lifetime of the install.
+    @objc public var deviceApplicationStamp: String? {
+        guard let userDefaults = MPUserDefaults.cached() else { return nil }
```
BrandonStalnaker, line 57: Nit: The class doc justifies the cached() == nil early-return for uniqueIdentifier (line 28), but deviceApplicationStamp has the same early-return unremarked. This is a behavior change from the ObjC version: MPUserDefaultsConnector.userDefaults created the singleton on first access and always minted+persisted a UUID, whereas cached() returns nil without minting. It's safe in practice as the singleton is created during MParticle init (_stateMachine = [[MPStateMachine_PRIVATE alloc] initWithUserDefaults:MPUserDefaultsConnector.userDefaults ...]) before any consumer-info read but I'd suggest a one-line comment here noting the pre-init behavior, mirroring the one above, so the difference is intentional-on-the-record.

BrandonStalnaker (dismissed):

mParticle-Apple-SDK-Swift/Sources/DataModel/MPConsumerInfo.swift
thomson-t: Good catch — agreed on both the asymmetry and that it is a real difference rather than a like-for-like port. Added in fdd69fbc.

Confirmed your reasoning before writing it down: `mParticle.m:186` builds the state machine inside `-[MParticle init]`, and `-[MPStateMachine_PRIVATE init]` reads `MPUserDefaultsConnector.userDefaults` (`MPStateMachine.m:50`), which goes through `+standardUserDefaultsWithConnector:` and creates the singleton — all before anything can reach consumer info. So the window is unreachable in practice.

The comment now records the pre-init behaviour explicitly and mirrors the one above it. I also noted why returning nil is the better answer than minting in that window: a stamp minted with nowhere to persist it would hand out a different value on the next call, which is worse than reporting none.

BrandonStalnaker (approved):

PR #971, opened 2026-09-08
refactor: inject backend persistence protocols

BrandonStalnaker:
## Background
- MPBackendController accesses persistence through global singleton state, making migration and focused testing difficult. Protocol injection is needed before migrating backend call paths.

## What Has Changed
- Added Objective-C-visible Swift persistence protocols.
- Injected persistence into MPBackendController.
- Migrated cleanup and retention operations to the injected Swift store.
- Updated backend tests to use protocol mocks.

## Checklist
- [x] I have performed a self-review of my own code.
- [x] I have made corresponding changes to the documentation.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] I have tested this locally.

## Reference Issue (For employees only. Ignore if you are an outside contributor)
- Closes https://go/j/[ticket-number]

PR #998, opened 2026-09-15
refactor: inject upload construction dependencies

thomson-t:
## Background

Upload construction reads SDK singletons directly, which prevents moving its implementation into the independent Swift module. This introduces the dependency boundary while preserving the Objective-C builder and keeping enrichment on the backend's injected persistence store.

## What Has Changed

- Add an Objective-C-callable upload context and enrichment persistence protocol with live providers.
- Inject identity, consent, options, fallback device/application information and logging at the backend boundary.
- Create upload contexts from the backend instance and resolve its current persistence at build time. Require enrichment operations through the backend persistence protocol.
- Cover injected-store isolation, forward-record deletion, store replacement after builder construction, build-time credentials/hooks and injected header values.
- Keep the main run loop available in session-end tests with asynchronous SDK queue work and bounded XCTest expectations.

## Screenshots/Video

N/A — no visual changes

## Checklist

- [ ] Self-review completed
- [ ] Tests added or updated
- [ ] Tested locally

## Additional Notes

Part 1 of the five-PR upload migration stack. Base: `workstation/swift-migration`. Merge in order; the base now includes kit-projection PRs #995 and #996.

The injected-store regression failed against the previous implementation: batching read and deleted forward records from the SDK singleton instead of the backend's injected store. Both new regressions pass with this fix, including replacing persistence after builder construction. No public API or ABI-baseline changes.

Validation before the kit-projection base refresh with Xcode 26.6:

- ABI guard and Trunk pass; no new issues.
- All-three-podspec lint passes, including iOS/tvOS builds.
- Complete iOS suites: 926 Objective-C and 920 Swift tests pass.
- Complete tvOS 18.5 suites: 850 Objective-C and 914 Swift tests pass.
- After restacking, final coordinator-head tests pass: 98 iOS / 97 tvOS backend and builder tests, plus 30 Swift builder/coordinator tests on each platform.

The earlier `testAutomaticSessionEnd` failure reproduced as a CI timeout twice. Its main-thread-blocking test setup, and the equivalent `testBackgroundBlock` setup, now use asynchronous SDK queue work and bounded expectations. Both passed the preceding Xcode 16.4 CI run and the current local suites. That preceding CI run also hit the independently reproduced `MParticleTests.testInitStartsSessionSyncDisableAll` ordering flake.

Fresh CI is pending for the updated stack, including the pinned Xcode 16.4 native jobs and Xcode 26.2 SwiftPM/WireMock integration job. The original 15 workspace edits are excluded.

Base refresh: the stack now includes `workstation/swift-migration` at `3e39aefc` (#995/#996). Xcode project conflicts were resolved by preserving both Objective-C deletions and both Swift test membership lists. On final stack head `fae4743b`, Trunk, ABI checks, and all-three-podspec lint pass; iOS passes 910 Objective-C / 973 Swift tests and tvOS passes 967 Swift tests. The tvOS Objective-C suite passes 833/834 tests; the previously observed `testInitStartsSessionSyncDisableAll` ordering flake passes in isolation. Both updated session-end tests and the injected-persistence regressions pass on both platforms. Fresh CI is requested for the updated dependent heads.

Stack: [#998](https://github.com/mParticle/mparticle-apple-sdk/pull/998) → [#999](https://github.com/mParticle/mparticle-apple-sdk/pull/999) → [#1000](https://github.com/mParticle/mparticle-apple-sdk/pull/1000) → [#1001](https://github.com/mParticle/mparticle-apple-sdk/pull/1001) → [#1002](https://github.com/mParticle/mparticle-apple-sdk/pull/1002).

## Summary by CodeRabbit

* **Improvements**
  * Uploads now use the latest SDK state, persistence data, device and application details, consent settings, and identifiers when prepared.
  * Batch transformations can replace upload data or cancel a batch when no transformed data is returned.
  * Upload metadata now receives current timestamps and message identifiers at build time.

* **Bug Fixes**
  * Improved consistency of enrichment data across batch, user, commerce, and ramp uploads.
  * Background upload operations now complete asynchronously without blocking the main run loop.


mParticle-Apple-SDK/MPBackendController.m
```diff
@@ -213,6 +216,50 @@ - (MPMessageBuilderContext *)messageBuilderContext {
...
 
+// The upload builder cannot import the public Objective-C SDK. Keep identity, consent,
+// build macros and customer callbacks at this composition boundary, with live providers.
++ (MPUploadBuilderContext *)uploadBuilderContext {
+    MPUploadBuilderContext *context = [[MPUploadBuilderContext alloc]
+        initWithStateMachine:^{ return MParticle.sharedInstance.stateMachine; }
+        lifetimeValue:^NSNumber *(NSNumber *mpid) {
+            return [MPUserDefaultsConnector.userDefaults mpObjectForKey:kMPLifeTimeValueKey userId:mpid] ?: @0;
+        }
+        persistence:^{ return (id<MPUploadEnrichmentPersistence>)MParticle.sharedInstance.persistenceStore; }
```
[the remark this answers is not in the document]
thomson-t: Fixed in 2047514 (rebased from b59ecc7). `uploadBuilderContext` is now an instance method, and its persistence provider reads the owning backend's current `self.persistence` at build time. `MPBackendPersistence` also explicitly includes the enrichment protocol.

Added regressions verifying that batching reads enrichment and deletes forward records only through the injected store, leaves the singleton store untouched, and observes persistence replacement after builder construction. The store-isolation test failed before the fix; both regressions now pass.

Validated the fix locally with both test schemes on iOS/tvOS, ABI checks, Trunk, and all-three-podspec lint. The dependent PRs also carry the fix.

mParticle-Apple-SDK-Swift/Sources/Persistence/MPPersistenceProtocols.swift
```diff
@@ -53,6 +53,7 @@ import Foundation
     MPPersistenceLifecycle,
     MPSessionPersistence,
     MPMessagePersistence,
-    MPUploadPersistence {}
+    MPUploadPersistence,
+    MPUploadEnrichmentPersistence {}
```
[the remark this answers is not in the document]
thomson-t: Technically it is correct, but MPBackendPersistence was introduced specifically for this migration in [#971](https://github.com/mParticle/mparticle-apple-sdk/pull/971)
Persistence injection is declared internally, without a customer-facing configuration point. We’re retaining the explicit enrichment requirement so backend batching consistently uses its injected store.

UnitTests/ObjCTests/MPBaseTestCase.m
```diff
@@ -36,8 +36,19 @@ @implementation MPTestConnectorFactory
 
 @end
 
+@interface MPBackendController_PRIVATE (UploadBuilderTesting)
+- (MPUploadBuilderContext *)uploadBuilderContext;
+@end
+
 @implementation MPBaseTestCase
 
+- (MPUploadBuilderContext *)uploadBuilderContext {
```
BrandonStalnaker, line 45: This allocates a fresh MPBackendController_PRIVATE on every call, and initWithDelegate: is not a cheap constructor — it registers six observers on NSNotificationCenter (UIApplicationDidEnterBackground, WillEnterForeground, DidFinishLaunching, DidBecomeActive, kMPNetworkPerformanceMeasurementNotification, and the remote-notification token on iOS), allocates an MPNetworkCommunication_PRIVATE, an MPNotificationController_PRIVATE and an operation queue. dealloc only calls endUploadTimer, so nothing unregisters them.

That controller then stays alive for as long as the returned context does, because the persistence:^{ return self.persistence; } block in -uploadBuilderContext captures it strongly. With 13 call sites across MPUploadBuilderTests, MPBackendControllerTests and this file, each test that takes a context leaves a live shadow backend controller subscribed to app-lifecycle notifications, which will run handleApplicationDidBecomeActive:/handleApplicationDidEnterBackground: against the shared singleton state. That's a good candidate for the testAutomaticSessionEnd and testInitStartsSessionSyncDisableAll ordering flakes called out in the description.

Could this reuse the controller that already exists instead of building a second one — return [[MParticle sharedInstance].backendController uploadBuilderContext];? initWithDelegate: already calls [[MParticle sharedInstance] initializePersistence], so the helper is touching the singleton either way and the "tests do not start the SDK" comment doesn't really hold.

mParticle-Apple-SDK/MPBackendController.m
```diff
@@ -213,6 +216,50 @@ - (MPMessageBuilderContext *)messageBuilderContext {
...
+            MPConsentState *consent = [MPPersistenceUtilities effectiveConsentStateForMpid:mpid];
+            return consent ? [MPConsentSerialization serverDictionaryFromConsentState:consent] : nil;
+        }
+        transformBatch:^id(NSDictionary *batch) {
+            MParticle *mparticle = MParticle.sharedInstance;
+            return mparticle.options.onCreateBatch ? mparticle.options.onCreateBatch(batch) : batch;
+        }
+        logger:^{ return [MParticle.sharedInstance getLogger]; }
+        sdkVersion:kMParticleSDKVersion];
+    context.timestamp = ^{ return MPMilliseconds([[NSDate date] timeIntervalSince1970]); };
```
BrandonStalnaker: Nit: MPMilliseconds(x) expands to @(trunc((x) * 1000)) (MPIConstants.h:6), which is exactly what MPUploadBuilderContext.timestamp already defaults to. This override is a second copy of the same conversion that can silently drift from the default. Worth deleting and letting the default stand — that would also make it consistent with messageID, which you do leave to its UUID().uuidString default.

BrandonStalnaker (commented):
Just one real issue I noticed

mParticle-Apple-SDK/MPBackendController.m
thomson-t: Thanks, removed the override.

UnitTests/ObjCTests/MPBaseTestCase.m
thomson-t: Agreed. The helper shouldn’t create another lifecycle-observing backend. Updated.

mParticle-Apple-SDK/MPBackendController.m
```diff
@@ -487,7 +534,7 @@ - (void)prepareBatchesForUpload:(MPUploadSettings *)uploadSettings {
                                                                     uploadInterval:self.uploadInterval
                                                                         dataPlanId:group.dataPlanId
                                                                    dataPlanVersion:group.dataPlanVersion
-                                                                    uploadSettings:uploadSettings];
+                                                                    uploadSettings:uploadSettings context:[self uploadBuilderContext]];
```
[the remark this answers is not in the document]
thomson-t: Yes, context creation now happens once per preparation call, outside both loops. I also carried this through the Swift coordinator migration in the downstream stack.

BrandonStalnaker (approved):

PR #1015, opened 2026-09-18
refactor: assemble crash and error messages in Swift

BrandonStalnaker:
## Background

- `MPBackendController.m` is the largest Objective-C file still in the migration's scope at 2,034 lines. Of those, `logError:` and `logCrash:` were 108 lines of dense message-dictionary assembly with no dispatch queues, no `@synchronized`, no main-thread requirement and no build macros — the kind of pure logic the migration is meant to move.
- Every collaborator the two methods need already lives in the Swift module: `MPMessageBuilder`, `MPBackendMessageInfo` (which already owns `base64CrashReport` and `crashReportBytesToRetain`), `MPBreadcrumb`, `MPMessage`, `MPSession`, `MPApplication_PRIVATE.appImageInfo()` and `MPPersistenceSchemaPRIVATE.maxBytesPerEvent(for:)`. Nothing new has to cross the boundary for the assembly to move.
- The cluster reaches `MParticle.sharedInstance` at exactly one point, `stateMachine.crashMaxPLReportLength`, which is an `NSNumber` and passes as a plain value — so the no-ObjC-import constraint is satisfiable here.
- `logCrash:` is protected by seven characterization tests that pin the payload to exact values. `logError:` has no direct test anywhere in the repository today.

## What Has Changed

- Adds `MPCrashMessageFactory` in `mParticle-Apple-SDK-Swift/Sources/Utils/`, with `crashMessage(...)` for the fatal report and `errorMessage(...)` for the handled error or exception.
- Keeps the factory pure, matching the contract `MPBackendMessageInfo` states alongside it: the caller reads persistence and writes the result. The breadcrumb trail and the recoverable-session list are passed in, and neither method saves.
- Preserves three behaviours that read like inconsistencies but are load-bearing: the crash is attributed to the first recoverable session that is not the current one (and to no session when there is none); the breadcrumb trail rides along with an exception only, never with a bare message; and the retained report length is measured against the assembled message, so truncation runs after the build rather than against the report alone.
- Mirrors the seven crash and error message keys in `MessageKeys.swift`. The Objective-C constants are untouched — they have external linkage and other callers.
- Adds `MPCrashMessageFactoryTests` with 28 tests covering severity and handled-ness, the exception and bare-message paths, breadcrumb presence, absence and undecodable entries, crash-session selection including every degenerate case, report capping and truncation, the topmost-context class name, event-info handling and the app-image merge. This is the first direct coverage the error path has had.
- Registers the new test file in both `membershipExceptions` lists in `project.pbxproj`.
- Changes no Objective-C. Nothing calls the new type yet; the cutover is the follow-up PR.

## Summary by CodeRabbit

- **New Features**
  - Added structured crash and handled-error reporting.
  - Crash reports can include severity, crash type, handled status, error messages, stack traces, context, and diagnostic report data.
  - Added breadcrumb information to provide additional details about events leading up to a crash or error.
  - Reports now associate events with the appropriate application session and include relevant application metadata.
  - Diagnostic data is encoded and size-limited to support reliable message delivery.


BrandonStalnaker:
closing in favor of https://github.com/mParticle/mparticle-apple-sdk/pull/1018

